PR #13, opened 2016-02-10
Streamed storage

Rhaea:


GuutBoy:
Looks ok to me. However, so many files are affected it is hard for me to check everything.

tpj:
Would be nice if pull request was open for a little longer than one hour, so I could get the chance to add comments too :-)

Rhaea:
Sure thing - however, you're most welcome to still add comments. 
Also, sometimes it is not possible to wait - but this time it was I guess :)

PR #14, opened 2016-02-11
Name change of DotProduct

Rhaea:
Changed name of dot-product to entrywise product. Also made clarifying descriptions to both entrywise product and inner product and factories


Rhaea:
Very non-damaging merge, so doing it now to be able to move on. Please still give comments if there are any.

PR #26, opened 2016-06-07
fixed source code example in docs

rdragos:
PS: It seems that the doc website is a little outdated.


GuutBoy:
Hi @rdragos , thanks for the update! It's true, there has been some changes since the doc site was written including a pretty large refactoring of names. We should really update it, thank you for bringing attention to the issue.

PR #33, opened 2016-08-16
Fix commandline utility

GuutBoy:
Fixes problems pointed out on the mailing list in the command line utility where configurations of most protocol suites was not done correctly. 

Additionally: 

- adds a simple demo application using the commandline utility. The demo application simply computes the distance between two points in two-dimensional space given by the parties. 

- fixes a problem that could cause tests to be run in parallel.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,9 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <version>2.18.1</version>
             <configuration>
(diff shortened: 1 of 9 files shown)


Rhaea:
Nice - This is great. The only thing I can think of is to include this new demo application in the documentation. 
Merge approved for my side.

GuutBoy:
I was exactly thinking that this would be a better example than the demo application currently in the documentation. In general the documentation needs a lot of work by now though ... need to find the time somehow :).

PR #44, opened 2016-11-14
Feature/builder

Rhaea:
Implements a master builder. 
This should make it easier to avoid problems when using multiple builders at the same time.


src/main/java/dk/alexandra/fresco/lib/helper/builder/AbstractProtocolBuilder.java
```diff
@@ -146,7 +177,19 @@ public ProtocolProducer getProtocol() {
...
+		}
 		curpp = new SequentialProtocolProducer();
 		producerStack = new Stack<AppendableProtocolProducer>();
 	}
+	
+	/**
+	 * Sets the parent of this builder meaning that the internal stack of the builder will be the parents. 
+	 * @param parent
+	 */
+	protected void setParentBuilder(AbstractProtocolBuilder parent){
```
Rhaea: This could be made public, granting the power of making any builder into a master builder. This would then support the style of coding that we currently have, while still avoiding problems with multiple builders.

src/main/java/dk/alexandra/fresco/lib/helper/builder/AbstractProtocolBuilder.java
```diff
@@ -67,6 +71,10 @@
 	 */
 	@Override
 	public void beginParScope() {
```
GuutBoy: Wouldn't it be cleaner to do 
`if (parent = null) {
  ParallelProtocolProducer ppp = new ParallelProtocolProducer();
  append(ppp);
  pushProducer(ppp);
} else {
  parent.beginParScopt(); 
}`

Rhaea: Well yea I guess. You'd eliminate the return statement which, if forgotten, would be bad.

src/main/java/dk/alexandra/fresco/lib/helper/builder/AbstractProtocolBuilder.java
Rhaea: Such that a functional builder needs a master to function? 

Sure, we could investigate that - then the question would be which builders are structural. The Omnibuilder for sure, but other than that right now, I don't know.

GuutBoy: You would only have a single structural builder. It could be the OmniBuilder, but as we already talked about I am not sure I like the idea of the OmniBuilder to begin with ... the point is that it is really messy that you can control scope using the sub builders provided by the OmniBuilder.

src/main/java/dk/alexandra/fresco/lib/helper/builder/AbstractProtocolBuilder.java
GuutBoy: Also, it would make the code easier to read. I really think we should do it that way.

PR #53, opened 2017-01-27
Lpsolving

MichaelStausholm:
The DEASolver should be ready to be merged.


src/main/java/dk/alexandra/fresco/lib/statistics/DEASolver.java
```diff
@@ -0,0 +1,246 @@
...
+public class DEASolver implements Application{
+
+	private static final long serialVersionUID = 7679664125131997196L;
+	private List<List<SInt>> targetInputs, targetOutputs;  
+	private List<List<SInt>> inputDataSet, outputDataSet; 
+	
+	private SInt[] optimal;
+	private OInt[] output;
+	
+	public DEASolver(
```
Rhaea: Would love a bit more context/documentation here. What are the different things, and what should these be. 
But cool application :)

PR #59, opened 2017-02-15
Modified the fuelstation to use only in-memory. 

Rhaea:
File usage would blow out most harddrives quickly at extended use, so with this we move away from file-based ways and towards keeping a small amount in memory at all times. 
We should still look into ways of streaming the data instead of getting it in chucks as is the case for now.


Rhaea:
I now believe it's ready for actual merging. I intent to do no more work except maybe more javadoc.

GuutBoy:
If you did not finish the javadoc, maybe you should do that before merge :).

Rhaea:
Should be ready for merge now

PR #63, opened 2017-02-22
Feature/division builder

markspanbroek:
Added a division builder to simplify the creation of division protocols.

### DIFF
--- a/src/main/java/dk/alexandra/fresco/lib/helper/builder/AdvancedNumericBuilder.java
+++ b/src/main/java/dk/alexandra/fresco/lib/helper/builder/AdvancedNumericBuilder.java
@@ -5,12 +5,12 @@ import dk.alexandra.fresco.framework.value.SInt;
 import dk.alexandra.fresco.framework.value.SIntFactory;
 import dk.alexandra.fresco.lib.math.integer.division.DivisionFactory;
 
-public class DivisionBuilder extends AbstractProtocolBuilder {
+public class AdvancedNumericBuilder extends AbstractProtocolBuilder {
 
     private DivisionFactory divisionFactory;
     private SIntFactory intFactory;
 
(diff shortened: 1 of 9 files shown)


Rhaea:
Not sure if we want a single builder for just division. Maybe more something like AdvancedNumericBuilder. Not entirely sure what else to put in that right now, but I'm sure more features will emerge. So do a renaming for now I guess, and I, for one, will be happy. 
Like the change with the abstract class as well! Makes sense

markspanbroek:
Good suggestion @Rhaea, I've renamed the class.

src/main/java/dk/alexandra/fresco/lib/helper/builder/AdvancedNumericBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private DivisionFactory divisionFactory;
+    private SIntFactory intFactory;
+
+    public AdvancedNumericBuilder(DivisionFactory divisionFactory, SIntFactory intFactory) {
+        this.divisionFactory = divisionFactory;
+        this.intFactory = intFactory;
+    }
+
+    public SInt div(SInt p, int maxBitLengthP, SInt q, int maxBitLengthQ) {
```
Rhaea: Oh yea.. two things: javadoc missing. One does not know if p/q or q/p. And maybe call them divident and divisor. 
And second thing: We would like methods for all the possible functionality that the division factory provides. So also division by known divisor and division where the remainder is returned as well etc.

markspanbroek:
Ok, I've added javadoc and included the other operations from the division factory. Could you please merge this pull request @Rhaea? Then I can create the next pull request (an improved division algorithm) which depends on the changes in this one.

Rhaea:
The policy is to have at least two reviewers look at a pull request before merging. This should happen soon though - I pinged Peter today.

PR #70, opened 2017-03-14
Feature/kryo newnetwork

Rhaea:
Implements a different network architecture which speeds up the current Scapi network and introduces a new network using the KryoNet framework. 

All tests go through and an Amazon AWS test was also done using the LP solver in SPDZ, so it definitely works. 

Lacking are:
- Thorough check of the homemade encrypted channel.
- Authenticated channels.
- Ability to choose the wanted network (KryoNet is hardcoded currently, as it is by far the fastest)


Rhaea:
Fixed the ability to choose network - this is now done from the SCEConfiguration

PR #117, opened 2017-06-30
Creating resource pools specific protocol suites and rewrite of composite protocols to be based on builders

pffrandsen:


core/src/main/java/dk/alexandra/fresco/framework/sce/configuration/ProtocolSuiteConfiguration.java
```diff
@@ -27,16 +27,37 @@
...
 
-  ProtocolSuite createProtocolSuite(int myPlayerId);
+  /**
+   * Legacy method for creating resource pool from an oblivious context - normally users of the
+   * SecureComputationEngine would control resources themselves.
+   * <br/>
+   * Once this methods is no longer needed, this interface will be removed
+   */
 
+  ResourcePoolT createResourcePool(
```
Rhaea: Add @Deprecated if it is deprecated

PR #123, opened 2017-07-14
Fix demos

Rhaea:
All but one demonstrator now works. The Aggregation demo was previously found within the demo directory, but even then was not working. Still seems like a relevant demonstrator, but it needs to be fixed. 
Merged with develop to add dummy protocol suites to all demonstrators.


GuutBoy:
Looks good. There are some formatting issues, mostly I noted imports are sorted out of style (https://google.github.io/styleguide/javaguide.html#s3.3-import-statements). You may need to adjust your IDE's formatter (I had similar issues so let me know if you need help setting it up). Also, I think we should organize the packages a bit different. Essentially put everything to do with demos in a subpackage of "dk.alexandra.fresco.demo". I.e., including the commons stuff.

PR #137, opened 2017-08-02
Bugfix/kryonet windows

Rhaea:
Should enable KryoNet to function on Windows.

### DIFF
--- a/core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
+++ b/core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
@@ -10,7 +10,6 @@ import dk.alexandra.fresco.framework.MPCException;
 import dk.alexandra.fresco.framework.configuration.NetworkConfiguration;
 import dk.alexandra.fresco.framework.crypto.AES;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
```diff
@@ -114,7 +114,9 @@ public NaiveListener(Map<Integer, BlockingQueue<byte[]>> queue) {
         if (pPort == null) {
           continue;
         }
-        if (pHost.equals(hostname) && pPort == port) {
+        // The special case of localhost == 127.0.0.1 enables windows users to play along.
+        if ((pHost.equals(hostname) || (pHost.equals("localhost") && hostname.equals("127.0.0.1")))
+            && pPort == port) {
           return queue.get(i);
```
pffrandsen: This has the disadvantage of not working in a complex network setup. The servers from BDbS have two network interfaces and host resolving is complicated for the application. Especially if it is run in a docker container.

Might I suggest a different strategy (or none at all) for checking the host name and party identifying? A simple version could just to forward your own contact details - given the security of the protocols this handshake only need to be correct, not unbreakable

PR #140, opened 2017-08-18
Removing calls to known

pffrandsen:
A lot (if not all) calls to known is actually just a way to bootstrap algorithms with a basic step. Introdcued these basic steps, hence improving performance ever so little.

As far as I can tell, one could reasonably deprecate known.


core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
```diff
@@ -46,12 +46,11 @@ public ExponentiationOpenExponent(Computation<SInt> x, BigInteger e) {
   @Override
   public Computation<SInt> build(SequentialNumericBuilder builder) {
     if (exponent.equals(BigInteger.ZERO)) {
-      return builder.numeric().known(BigInteger.valueOf(1));
+      throw new RuntimeException("Now why would you exponentiate a SInt with 0?");
```
GuutBoy: You may want to exponentiate with by a value not known at compile time which could be zero? Of course you could test for that outside of this class and mark it as only compatible with strictly positive exponents. However, in that case this should be an "illegal argument" exception.

GuutBoy: Additionally, the exception should probably be thrown for any value of the exponent that is not strictly positive.

pffrandsen: Agree, branch updated.

Given that this is the open exponentiate, you will now - however maybe not runtime.

GuutBoy: Possibly the error message could be less judgemental as well?

core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
```diff
@@ -41,13 +41,13 @@
   public ExponentiationOpenExponent(Computation<SInt> x, BigInteger e) {
     this.base = x;
     this.exponent = e;
+    if (exponent.compareTo(BigInteger.ZERO) <= 0) {
+      throw new IllegalArgumentException("Now why would you exponentiate a SInt with 0?");
```
GuutBoy: "... 0 or less"

core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
```diff
@@ -42,7 +42,8 @@ public ExponentiationOpenExponent(Computation<SInt> x, BigInteger e) {
     this.base = x;
     this.exponent = e;
     if (exponent.compareTo(BigInteger.ZERO) <= 0) {
-      throw new IllegalArgumentException("Now why would you exponentiate a SInt with 0?");
+      throw new IllegalArgumentException(
+          "Now why would you exponentiate a SInt with somthing equal to or less than 0?");
```
GuutBoy: "something"

core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
```diff
@@ -43,7 +43,7 @@ public ExponentiationOpenExponent(Computation<SInt> x, BigInteger e) {
     this.exponent = e;
     if (exponent.compareTo(BigInteger.ZERO) <= 0) {
       throw new IllegalArgumentException(
-          "Now why would you exponentiate a SInt with somthing equal to or less than 0?");
+          "This computation does not support exponentiate a SInt with somthing equal to or less than 0?");
```
GuutBoy: something

pffrandsen: Changed the wording

PR #143, opened 2017-08-24
Dummy booleans

### LINKED ISSUE #107: Make the TinyTables protocol suite compatible with new framework structure
Make TinyTables compatible with the new structure similar to the SPDZ protocol suite

### LINKED ISSUE #110: Update the Demo project to be compatible with the new framework design
The Demo's in the demo project are not up to date with the recent framework updates. This should be fixed.

EDIT: This includes using the new style of protocol creation using the ProtocolBuilder instead of the old ProtocolFactories.

### LINKED ISSUE #130: Update tests to be compatible with the new framework design
The tests needs to be updated to be compatible with the new framework design. I.e., they should all use the new style of protocol creation using the ProtocolBuilders instead of the old ProtocolFactories.

### LINKED ISSUE #131: Remove the ProtocolFactories
The old ProtocolFactories remain only to support old code using them. Once we have updated this code to use the ProtocolBuilders instead we should get rid of the ProtocolFactories entirely, so that only the ProtocolBuilders are used. 

This depends on first solving #130 #110 and possibly #111.

Rhaea:
First of all: Very large PR - sry. 

We changed and molded the boolean MPC stuff into the new architecture. This includes creating the same lambda style as the numeric parts, changing all used binary protocols, reworked the dummy boolean suite as well as tinytables and finally fixed the demos to match the new arch. All binary tests are also written in the new style. 

On top of this, Michael also brought the test coverage up from not very much, to a lot :)

Closes #131 
Closes #130 
Closes #110 
Closes #107


core/src/main/java/dk/alexandra/fresco/framework/value/SIntFactory.java
```diff
@@ -45,10 +55,9 @@
   SInt getSInt(BigInteger i);
 
   /**
-   * Gets a protocol to load a publicly known value into a SInt.
-   * The idea here is to not do the computation involved in loading the SInt
-   * while we are building the protocol.
-   * TODO: This should not be how values are loaded
+   * Gets a protocol to load a publicly known value into a SInt. The idea here is to not do the
+   * computation involved in loading the SInt while we are building the protocol. TODO: This should
+   * not be how values are loaded
```
pffrandsen: This realignment is very confusing since it obscures the TODO

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMergeProtocolRec.java
```diff
@@ -66,7 +64,7 @@
   private final int realSize;
   private final int simulatedSize;
   private int layers;
-  private List<Layer> layerList;
+  //private List<Layer> layerList;
```
pffrandsen: Remove?

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMergeProtocolRec.java
```diff
@@ -79,12 +77,11 @@
    */
   public OddEvenMergeProtocolRec(List<Pair<SBool[], SBool[]>> left,
       List<Pair<SBool[], SBool[]>> right,
-      List<Pair<SBool[], SBool[]>> sorted, AbstractBinaryFactory factory) {
+      List<Pair<SBool[], SBool[]>> sorted, Object factory) {
```
pffrandsen: Remove factory?

core/src/main/java/dk/alexandra/fresco/lib/compare/CompareAndSwap.java
```diff
@@ -26,26 +26,27 @@
...
-    CompareAndSwapProtocol {
 
-	private SBool[] left;
-	private SBool[] right;
-	private AbstractBinaryFactory bp;
+public class CompareAndSwap implements ComputationBuilder<List<SBool>> {
+
+	private Computation<List<SBool>> left;
+	private Computation<List<SBool>> right;
+	//private AbstractBinaryFactory bp;
```
pffrandsen: Remove

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/BinaryGreaterThanProtocolImpl.java
```diff
@@ -3,115 +3,80 @@
...
-	}
+  @Override
+  public Computation<SBool> build(SequentialBinaryBuilder builder) {
+    return builder.seq(seq -> {
+      int round = 0;
+      Computation<SBool> xor = seq.binary().xor(inA.get(length - 1), inB.get(length - 1));
+      round++;
+      Computation<SBool> outC = seq.binary().and(inA.get(length - 1), xor);
+      round++;
+      for (; round <= length; round++) {
```
pffrandsen: This would be chaeper in terms of memory if done via whileLoop

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/eq/BinaryEqualityProtocolImpl.java
```diff
@@ -3,119 +3,85 @@
...
+      return () -> xnors;
+    }).seq((xnors, seq) -> {
+      // xnors are now a bitstring where a 0 represents that something differed between the inputs
+      // Can now do an AND row to determine if the entire bitstring is 1's (inputs are equal) or
+      // not.
+      // TODO: One can obtain better results using a tree structure and doing those in parallel.
+      Computation<SBool> xnorsAnd;
+      xnorsAnd = seq.binary().and(xnors.get(0), xnors.get(1));
+      int i = 2;
+      while (i < length) {
```
pffrandsen: Consider using the whileLoop construct here

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/eq/BinaryEqualityProtocolImpl.java
```diff
@@ -3,119 +3,85 @@
...
-              xnorOuts[i + step], xnorOuts[i]);
-          parAND.append(and);
-          i += 2 * step;
-        }
-        curPP = parAND;
+      return () -> xors;
+    }).par((xors, par) -> {
+      List<Computation<SBool>> xnors = new ArrayList<>();
+      for (int i = 0; i < length; i++) {
+        xnors.add(par.binary().not(xors.get(i)));
```
pffrandsen: This seems to be equivalent to doing xnor in the par.advanced()

Either remove that method or use it here?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -121,145 +112,127 @@ private boolean checkDimensions(LPTableau tableau, Matrix<Computation<SInt>> upd
...
-            if (!phaseOneOutput.terminated()) {
-              phaseTwoProtocol(inner, phaseOneOutput);
-            }
-            return phaseOneOutput;
-          });
-        }).seq((whileState, seq) ->
-        () -> new LPOutput(whileState.tableau, whileState.updateMatrix, whileState.basis,
-            whileState.pivot)
-    );
+    }).whileLoop(state -> !state.terminated(), (state, seq) -> {
```
pffrandsen: This is not a good realignment, now the condition is obscured by the lambda construction

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -121,145 +112,127 @@ private boolean checkDimensions(LPTableau tableau, Matrix<Computation<SInt>> upd
...
-    }).seq((pair, seq) -> {
-      List<Computation<SInt>> basis = pair.getFirst();
-      state.updateMatrix = pair.getSecond();
-      state.basis = basis;
-//      // Copy the resulting new update matrix to overwrite the current
-      state.prevPivot = state.pivot;
-      return () -> state;
-    });
+  private Computation<LPState> phaseTwoProtocol(SequentialNumericBuilder builder, LPState state) {
+    return builder.seq((seq) -> seq.createSequentialSub(
```
pffrandsen: This is also a bad realignment

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -121,145 +112,127 @@ private boolean checkDimensions(LPTableau tableau, Matrix<Computation<SInt>> upd
...
-      List<Computation<SInt>> entering = enteringAndMinimum.getFirst();
-      SInt minimum = enteringAndMinimum.getSecond();
-      // Check if the entry in F is non-negative
-      Computation<SInt> positive = seq.comparison().compareLEQLong(zero, () -> minimum);
-      state.terminationOut = seq.numeric().open(positive);
-      state.enteringIndex = entering;
-      return () -> state;
-    });
+    return builder
+        .seq(
```
pffrandsen: Strange newline?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -121,145 +112,127 @@ private boolean checkDimensions(LPTableau tableau, Matrix<Computation<SInt>> upd
...
-      List<Computation<SInt>> entering = enteringAndMinimum.getFirst();
-      SInt termination = enteringAndMinimum.getSecond();
-      state.terminationOut = seq.numeric().open(() -> termination);
-      state.enteringIndex = entering;
-      return () -> state;
-    });
+  private Computation<LPState> blandPhaseOneProtocol(SequentialNumericBuilder builder,
+      LPState state) {
+    return builder
+        .seq(
```
pffrandsen: ditto

core/src/main/java/dk/alexandra/fresco/lib/lp/LPTableau.java
```diff
@@ -49,7 +46,7 @@ public LPTableau(Matrix<Computation<SInt>> C, ArrayList<Computation<SInt>> B,
       this.F = F;
       this.z = z;
     } else {
-      throw new RuntimeException("Dimenssions of tableau does not match");
+      throw new IllegalArgumentException("Dimenssions of tableau does not match");
```
pffrandsen: While you are at it, fix the spelling error

core/src/main/java/dk/alexandra/fresco/lib/math/integer/binary/RightShift.java
```diff
@@ -49,12 +46,9 @@
...
    * @param input The input.
    * @param calculateRemainders true to also calculate remainder. If false remainders in result will
-   * be null.
+   *        be null.
    */
-  public RightShift(
-      int bitLength,
-      Computation<SInt> input,
-      boolean calculateRemainders) {
+  public RightShift(int bitLength, Computation<SInt> input, boolean calculateRemainders) {
```
pffrandsen: I have a hard time tracking the changes here - is it just realignment?

core/src/main/java/dk/alexandra/fresco/suite/dummy/arithmetic/DummyArithmeticSInt.java
```diff
@@ -92,4 +92,23 @@ public String toString() {
   public SInt out() {
     return this;
   }
+
+  @Override
+  public boolean equals(Object obj) {
```
pffrandsen: Equals implies hashcode as well - add both or none

core/src/main/java/dk/alexandra/fresco/suite/dummy/arithmetic/DummyArithmeticFactory.java
```diff
@@ -31,10 +31,12 @@
 import dk.alexandra.fresco.lib.field.integer.BasicNumericFactory;
 import java.math.BigInteger;
 
+public class DummyArithmeticFactory implements BasicNumericFactory{
```
pffrandsen: Realignment removed the javadoc for the class?

core/src/main/java/dk/alexandra/fresco/suite/dummy/arithmetic/DummyArithmeticFactory.java
```diff
@@ -60,6 +62,15 @@ public SInt getSInt(BigInteger i) {
   }
 
   @Override
+
+  public SInt[] getSIntArray(int amount) {
```
pffrandsen: See interface - this should not be added

core/src/test/java/dk/alexandra/fresco/lib/math/bool/mult/TestMult.java
```diff
@@ -0,0 +1,19 @@
+package dk.alexandra.fresco.lib.math.bool.mult;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import dk.alexandra.fresco.framework.value.SBool;
+
+public class TestMult {
+
+  @Ignore
```
pffrandsen: Why include this ignored test?

core/src/test/java/dk/alexandra/fresco/lib/math/bool/mult/TestMult.java
```diff
@@ -0,0 +1,19 @@
...
+  @Test(expected = IllegalArgumentException.class)
+  public void testInconsistentLengthFullAdder2(){
+    SBool[] left = new SBool[8];
+    SBool[] right = new SBool[8];
+    SBool[] output = new SBool[18];
+    //fact.getBinaryMultProtocol(left, right, output);
```
pffrandsen: Because?

core/src/main/java/dk/alexandra/fresco/framework/builder/NumericBuilder.java
```diff
@@ -31,6 +33,14 @@
 
   Computation<SInt> known(BigInteger value);
 
+  default List<Computation<SInt>> known(List<BigInteger> value) {
```
pffrandsen: To the best of my knowledge known is just a lazy way of sharing constant input data. I still have not seen a place where known really is used for something essential. I recently pushed a branch to eliminate every call to known apart from LP/DEA and CompareToZero/sign (lacking a concrete implementation of this).

So instead of adding more usages to I would recommend removing it altogether - and at least not add more utlity functions for working with known

core/src/main/java/dk/alexandra/fresco/framework/builder/BuilderFactoryNumeric.java
```diff
@@ -37,4 +40,8 @@ default AdvancedNumericBuilder createAdvancedNumeric(ProtocolBuilderNumeric buil
   default SequentialNumericBuilder createProtocolBuilder() {
     return ProtocolBuilderNumeric.createApplicationRoot(this);
   }
+
+  default UtilityBuilder createUtilityBuilder(ProtocolBuilderNumeric builder) {
```
pffrandsen: UtilityBuilder is not necessarily the best wording - utlity is very generic, and this interface has a very specific purpose. Maybe DebugBuilder or DebugViaOpenBuilder

Actually it is not needed to make as a builder, it consist of four different ComputationBuilders that can be inlined whereever used. No need (nor reason) to make it a part of the BuilderFactoryNumeric interface

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
```diff
@@ -73,30 +69,29 @@ public SecureComputationEngineImpl(
...
+  public <OutputT> OutputT runApplication(Application<OutputT, Builder> application,
+      ResourcePoolT sceNetwork) {
     Future<OutputT> future = startApplication(application, sceNetwork);
     try {
       return future.get(10, TimeUnit.MINUTES);
     } catch (InterruptedException | TimeoutException e) {
       throw new RuntimeException("Internal error in waiting", e);
     } catch (ExecutionException e) {
-      throw new RuntimeException("Execution exception when running the application", e.getCause());
+      throw new RuntimeException("Execution exception when running the application", e);
```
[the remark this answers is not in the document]
Rhaea: The reason behind this was to be able to reach the cause itself, making it possible to check within tests which exception caused the failure. It is way easier to do exception testing when you can reach the cause, so I still hold on to keeping this change.

core/src/main/java/dk/alexandra/fresco/framework/builder/binary/BinaryBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+ * Interface for the basic operations which any binary protocol suite needs to implement.
+ * 
+ * @author Kasper Damgaard
+ *
+ */
+public interface BinaryBuilder {
+
+  Computation<SBool> known(boolean known);
+
+  default List<Computation<SBool>> known(boolean[] known) {
```
pffrandsen: Known should be used for fixed constants in the calculations. Returning a list is weird, since boolean arithmetic only has two constants. 

In this case it is just testing that is made easy. I would suggest either to use input in those test, move this implementation to test or even stream the arrays (comp1.stream().map(builder::known).collect(...) )

core/src/main/java/dk/alexandra/fresco/framework/builder/binary/BinaryUtilityBuilder.java
```diff
@@ -0,0 +1,41 @@
+package dk.alexandra.fresco.framework.builder.binary;
+
+import dk.alexandra.fresco.framework.Computation;
+import dk.alexandra.fresco.framework.value.SBool;
+import java.io.PrintStream;
+import java.util.List;
+
+public interface BinaryUtilityBuilder {
```
pffrandsen: See similar comment about numeric utility builder - this interface ought not to exist

core/src/main/java/dk/alexandra/fresco/framework/builder/binary/BuilderFactoryBinary.java
```diff
@@ -0,0 +1,36 @@
...
+
+  default BinaryBuilderAdvanced createAdvancedBinary(ProtocolBuilderBinary builder) {
+    return new DefaultBinaryBuilderAdvanced(builder);
+  }
+
+  default BristolCryptoBuilder createBristolCryptoBuilder(ProtocolBuilderBinary builder) {
+    return new DefaultBristolCryptoBuilder(builder);
+  }
+
+  default BinaryUtilityBuilder createUtilityBuilder(ProtocolBuilderBinary builder) {
```
pffrandsen: See comment on interface description

core/src/main/java/dk/alexandra/fresco/framework/builder/binary/ComparisonBuilderBinary.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * Performs a greater than operation on the two inputted bitstrings which each represents a number
+   * in base 2.
+   * 
+   * @param inLeft The left secret shared bitstring
+   * @param inRight The right secret shared bitstring
+   * @return inLeft > inRight
+   */
+  public Computation<SBool> greaterThan(List<Computation<SBool>> inLeft,
```
pffrandsen: public?

core/src/main/java/dk/alexandra/fresco/framework/builder/binary/ComparisonBuilderBinary.java
```diff
@@ -0,0 +1,29 @@
...
+      List<Computation<SBool>> inRight);
+
+  /**
+   * Performs an equality operation on the two inputted bitstrings.
+   * 
+   * @param inLeft The left secret shared bitstring
+   * @param inRight The right secret shared bitstring
+   * @return inLeft == inRight
+   */
+  public Computation<SBool> equal(List<Computation<SBool>> inLeft,
```
pffrandsen: Interfaces are by default public - hence visibility is redundant

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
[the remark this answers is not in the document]
pffrandsen: Normally you would never need this. So could you point to the test in question that requires it?

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMergeProtocolImpl.java
```diff
@@ -26,133 +26,140 @@
...
-		blb.endCurScope();
-		initializeIndices();
-		newMerge(0, simulatedSize, 1);
-		return blb.getProtocol();
-	}
+  @Override
+  public Computation<List<Pair<List<Computation<SBool>>, List<Computation<SBool>>>>> build(
+      ProtocolBuilderBinary builder) {
+    // TODO Auto-generated method stub
+    return null;
```
pffrandsen: This seems unhealthy - could we complete the refactor here (or remove it all together)

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMergeProtocolRec.java
```diff
@@ -114,7 +111,7 @@ public OddEvenMergeProtocolRec(List<Pair<SBool[], SBool[]>> left,
 
   @Override
   protected ProtocolProducer initializeProtocolProducer() {
-    BasicLogicBuilder blb = new BasicLogicBuilder(factory);
+    /*BasicLogicBuilder blb = new BasicLogicBuilder(factory);
```
pffrandsen: This seems quite unhealthy

core/src/main/java/dk/alexandra/fresco/lib/compare/CompareAndSwap.java
```diff
@@ -64,5 +61,11 @@ protected ProtocolProducer initializeProtocolProducer() {
 
 		blb.endCurScope();
 		return blb.getProtocol();
-	}
+	}*/
+
+  @Override
+	public Computation<List<SBool>> build(ProtocolBuilderNumeric builder) {
+		// TODO Auto-generated method stub
+    return null;
```
pffrandsen: What happened to the code - this should be completed?

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/ParBinaryGreaterThanProtocolImpl.java
```diff
@@ -104,18 +100,18 @@ public ProtocolProducer nextProtocolProducer() {
 		} else {
 			return null;
 		}
-		return blb.getProtocol();
+		return blb.getProtocol();*/ return null;
```
pffrandsen: null will break the computation - should this be removed?

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/eq/AltBinaryEqualityProtocol.java
```diff
@@ -80,16 +70,17 @@ public ProtocolProducer nextProtocolProducer() {
 			xnorOuts[0] = out;
 			blb.addProtocolProducer(new TreeProtocol(this));
 			round++;
-		} else if (round == 3){
+		} else {*/
 			return null;
```
pffrandsen: null is dangerous - it will not function, maybe remove?

core/src/main/java/dk/alexandra/fresco/lib/debug/ArithmeticOpenAndPrint.java
```diff
@@ -0,0 +1,109 @@
...
+  public ArithmeticOpenAndPrint(String label, Matrix<Computation<SInt>> matrix,
+      PrintStream stream) {
+    this.label = label;
+    this.matrix = matrix;
+    this.stream = stream;
+  }
+
+  @Override
+  public Computation<Void> build(ProtocolBuilderNumeric builder) {
+    return builder.seq(seq -> {
```
pffrandsen: This would be a lot easier to read if split into three different classes

core/src/main/java/dk/alexandra/fresco/lib/field/bool/AndProtocolFactory.java
```diff
@@ -26,15 +26,14 @@
...
+import dk.alexandra.fresco.framework.Computation;
 import dk.alexandra.fresco.framework.ProtocolFactory;
-import dk.alexandra.fresco.framework.ProtocolProducer;
-import dk.alexandra.fresco.framework.value.OBool;
 import dk.alexandra.fresco.framework.value.SBool;
 
 public interface AndProtocolFactory extends ProtocolFactory {
 
-  ProtocolProducer getAndProtocol(SBool inLeft, SBool inRight, SBool out);
+  Computation<? extends SBool> getAndProtocol(SBool inLeft, SBool inRight, SBool out);
```
pffrandsen: This class is unused - and should be deleted

core/src/main/java/dk/alexandra/fresco/lib/field/bool/BasicLogicFactory.java
```diff
@@ -44,9 +43,9 @@
  * 
  *
  */
+@Deprecated
```
pffrandsen: Go through all the interface and remove+delete the unused interfaces and methods. As far as I can tell it is every one.

core/src/main/java/dk/alexandra/fresco/lib/field/bool/generic/AndFromCopyConst.java
```diff
@@ -24,12 +24,39 @@
...
+/**
+ * This protocol implements 
+ *
+ * a AND b
+ *
+ * as
+ *
+ * if (b) then a ELSE false 
+ */
+public class AndFromCopyConst implements
```
pffrandsen: Rename - copy does not apply anymore

core/src/main/java/dk/alexandra/fresco/lib/field/bool/generic/OrFromCopyConst.java
```diff
@@ -0,0 +1,57 @@
...
+
+import dk.alexandra.fresco.framework.Computation;
+import dk.alexandra.fresco.framework.builder.binary.ProtocolBuilderBinary;
+import dk.alexandra.fresco.framework.value.SBool;
+
+/**
+ * An efficient way of OR'ing an SBool with an OBool if we can construct SBools
+ * of constants and if we can copy protocols.
+ */
+public class OrFromCopyConst implements
```
pffrandsen: This computation should not copy (nor be called as such)

core/src/main/java/dk/alexandra/fresco/suite/dummy/bool/DummyBooleanAndProtocol.java
```diff
@@ -0,0 +1,87 @@
...
+   * 
+   * <p>
+   * Lets the caller specify where to store the output. This is for backward compatibility.
+   * </p>
+   * 
+   * @param left the left operand
+   * @param right the right operand
+   * @param out the {@link SBool} in which to store the output
+   */
+  public DummyBooleanAndProtocol(Computation<SBool> left, Computation<SBool> right, SBool out) {
```
pffrandsen: dead code - delete

core/src/main/java/dk/alexandra/fresco/lib/math/integer/binary/RightShift.java
Rhaea: Yea :)

core/src/main/java/dk/alexandra/fresco/framework/builder/BuilderFactoryNumeric.java
Rhaea: While this is true in theory, I suspect new users will never ever find these helpful debug methods. That is the main reason why I included it. 
I will, however, rename it to DebugBuilder and include javadoc to warn against using it in production.

core/src/main/java/dk/alexandra/fresco/lib/math/integer/binary/RightShift.java
pffrandsen: Agree - try to be careful with the realignments in the future. It resides in the commit "Initial work on binary version. Does not compile, but got a dummy test to run. So far so good." which is not accurate then.

core/src/main/java/dk/alexandra/fresco/framework/builder/BuilderFactoryNumeric.java
pffrandsen: I would still very much insist it is not available at all in production code.
How about leaving the class as is, i.e. leave the the artist formerly known as DefaultUtilityBuilder, but remove the interface from the BuilderFactoryNumeric. Then there is a perfectly fine debugger - but now it is not hardwired into the interface - but instantiable everywhere,.

core/src/main/java/dk/alexandra/fresco/framework/builder/NumericBuilder.java
Rhaea: Known is supposed to be used by application developers when they have a publicly known value which is used within the application. Most of our interfaces (if not all) does not support using just a BigInteger, but instead a Computation<SInt>. This means that if known is not present, one would have to input the value which requires communication. This is not preferred. 
I stick with keeping it. It also makes it easier to write tests instead of first having to input.

pffrandsen: First of all. The production API should not support easy testing that would be wrong.

There might be cases (especially in the boolean arithmetic) where the constants can be used in the computations. I am not avid supporter of removing the known per se. But the list of BigInteger is just lazy testers. If the is the desired way of writing Tests then by all means have it in the tests. But do not include it in the production code.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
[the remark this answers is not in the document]
Rhaea: ExponentiationTests.TestExponentiationZeroExponent

If you have some other way of checking if the correct exception is thrown, please let me know.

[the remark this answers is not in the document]
pffrandsen: Ah.

How about deleting the .getCause() both places?

It really is internal to the SecComEng that the stack is two deep.

Rhaea:
I'm very happy that you're fixing comments yourself, but maybe to avoid doing the same things twice, you could either wait with the comment until you fixed it, and say it's fixed, or state that you will do this soon? :)

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/ParBinaryGreaterThanProtocolImpl.java
Rhaea: I am deleting this since we never refer to it, and it's stated in the doc that it was worse than the simpler implementation

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/eq/AltBinaryEqualityProtocol.java
Rhaea: Deleting this too since we never refer to it, and if we ever want to optimize, we should do this on the primary implementation. No reason to have a slow one.

Rhaea: Oh, but if you have a good idea of how to do a tree based approach using the new arch, I would love to hear about this. (Maybe I'll just pop by your office tomorrow)

pffrandsen: Oh we do and are looking forward to your vist.

However, SumSIntList and ProductSIntList are both doing a tree based collation for 2-input gates

pffrandsen:
Sorry about the interleaving fixes - I solved another problem and just stepped across it.

pffrandsen:
A comment about the description: "... finally fixed the demos to match the new arch." 

Not excactly, see for instance DistanceDemo - here are two imports:
- import dk.alexandra.fresco.lib.helper.builder.NumericIOBuilder;
- import dk.alexandra.fresco.lib.helper.builder.NumericProtocolBuilder;

core/src/main/java/dk/alexandra/fresco/lib/compare/bool/eq/AltBinaryEqualityProtocol.java
MichaelStausholm: I actually already made the tree based protocol, it should be commit in a second.

core/src/main/java/dk/alexandra/fresco/suite/dummy/arithmetic/DummyArithmeticProtocolSuite.java
```diff
@@ -72,18 +65,21 @@ public int getMaxBitLength() {
...
-          throws IOException {}
+          throws IOException {
+      }
     };
   }
 
   @Override
   public DummyArithmeticResourcePool createResourcePool(int myId, int size, Network network,
       Random rand, SecureRandom secRand) {
-    return new DummyArithmeticResourcePoolImpl(myId, size, network, rand, secRand, modulus);
```
Rhaea: @pffrandsen : This breaks all tests which uses the dummy arithmetic suite. Reason being a null pointer from basicNumericContext which is set on init (which requires a resourcepool). That is a loop that never ends well :)
We need the createResourcePool to be independent of having init called or not.

Issue #34, opened 2016-09-09
How to run SPDZ applications


meilof:
Hi guys,

I was playing around with FRESCO and trying to run an application using the SPDZ backend. My command lines are:

java -jar lp.jar -i1 -p1:localhost:9292 -p2:localhost:9994 -s spdz
java -jar lp.jar -i2 -p1:localhost:9292 -p2:localhost:9994 -s spdz

However, it gives the following exception:

Exception in thread "main" java.lang.NullPointerException
        at dk.alexandra.fresco.suite.spdz.storage.DataSupplierImpl.getSSK(DataSupplierImpl.java:138)
        at dk.alexandra.fresco.suite.spdz.storage.SpdzStorageImpl.getSSK(SpdzStorageImpl.java:112)
        at dk.alexandra.fresco.suite.spdz.evaluation.strategy.SpdzProtocolSuite.init(SpdzProtocolSuite.java:133)
        at dk.alexandra.fresco.framework.sce.SCEImpl.setup(SCEImpl.java:169)
        at dk.alexandra.fresco.framework.sce.SCEImpl.runApplication(SCEImpl.java:232)
        at LPTest.main(LPTest.java:67)

Debugging says that indeed, the storage is set to NULL. But how can I change this?

meilof:
OK, I got things to work now by:

- Adding a command line option to select storage (which was not there in CmdLineUtil)
- Adding "-r STREAMED_STORAGE -D spdz.useDummyData=True" to the command line

However, it works only with dummy storage. The reason seems to be that the output of FakeTripleGen is incompatible with SpdzStorageImpl, is that correct?

GuutBoy:
Hi Meilof. Thank you for bringing this up. We are super busy right now. But I will look into this as soon as there is some time to breathe :).

Den 13. sep. 2016 kl. 11.03 skrev meilof <notifications@github.com<mailto:notifications@github.com>>:


OK, I got things to work now by:

  *   Adding a command line option to select storage (which was not there in CmdLineUtil)
  *   Adding "-r STREAMED_STORAGE -D spdz.useDummyData=True" to the command line

However, it works only with dummy storage. The reason seems to be that the output of FakeTripleGen is incompatible with SpdzStorageImpl, is that correct?

-

Rhaea:
Hey Meilof
[...]

PR #156, opened 2017-09-15
Spdz mac check issues

### LINKED ISSUE #152: Check the spdz maccheck protocol
Make sure the mac check protocol of spdz is used correctly

Rhaea:
This fixes #152 and #34. We no longer cut corners, but actually check for cheating before and after output is made.


Rhaea:
@pffrandsen I agree that indeed we could do a Mac check before outputs and then after outputs as long as it is in the same batch. This holds true. However, in order for me not to wreck the native protocol round concept, I need the output protocols to not be evaluated by the evaluator, but only by SPDZ itself. This is currently not supported, and is so far pretty specific to SPDZ. Therefore I did not change the design. 
I agree that prettier ways to do this is possible, but would require design changes that we need to ensure are not just there for SPDZ only. Most likely other protocol suites using the MAC style could benefit though which includes TinyTables once active security is there. 
I'll wait for the other Peter to respond as well before making design changes.

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/SpdzBatchedStrategy.java
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * @param protocols array holding the protocols to be evaluated
+   * @param sceNetwork array of sceNetworks corresponding to the protocols to be evaluated. I.e.,
+   *        the array should contain numProtocols SCENetworks, with sceNetwork[i] used for
+   *        communication in protocols[i].
+   * @param channel string indicating the channel to communicate over.
+   * @param rp the resource pool.
+   * @param startingRound The round to start from
+   */
```
GuutBoy: The documentation is not in sync with the method.

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/SpdzRoundSynchronization.java
```diff
@@ -4,55 +4,70 @@
...
  */
 public class SpdzRoundSynchronization implements RoundSynchronization<SpdzResourcePool> {
 
   private static final int macCheckThreshold = 100000;
+  private int gatesEvaluated = 0;
 
-  private void MACCheck(SpdzResourcePool resourcePool,
-      SCENetwork sceNetworks) throws IOException {
-
+  private void doMACCheck(SpdzResourcePool resourcePool, SCENetwork sceNetworks)
```
GuutBoy: sceNetworks -> sceNetwork

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzOutputProtocol.java
```diff
@@ -0,0 +1,15 @@
+package dk.alexandra.fresco.suite.spdz.datatypes;
```
GuutBoy: Shouldn't this by in spdz.gates?

pffrandsen:
The minor of the two issues has been fixed.

The major has three resolutions;
- We keep the current hack but change the rest of the code to match the new interface. I would be happy to share the very long list of entailed changes if we redefine the NativeProtocol interface. 
- We rework the current solution to use and work with round synchronization
- We live with the current hack

The latter is unacceptable in my opinion

GuutBoy:
Hej Peter,

Det kan der være noget om. Jeg satte den bare til default.

Jeg har været syg i går og er stadig meget dårlig. Jeg vil kigge på det når jeg er tilbage.

/Peter

Rhaea:
I actually liked the beforeBatch idea once I started pondering it. It makes it possible also as a protocol suite to inject or even remove protocols within the batch. In this case, we just check if an output gate exists, and then do a MAC check on all previous layers. We then finish the batch with another MAC check if an output gate existed in the current batch. 
I removed the hack.
All should be good now from my point of view.

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzOutputProtocol.java
```diff
@@ -1,95 +1,13 @@
...
-  }
+/**
+ * Marks a protocol as being able to output to some party. This means that the protocol should be
+ * added to the list of output protocols in the current batch. It also means that calling eval
+ * starting from round 1 will trigger the actual evaluation of the output protocol.
+ * 
+ * @author Kasper Damgaard
+ *
+ */
+public abstract class SpdzOutputProtocol<OutputT> extends SpdzNativeProtocol<OutputT> {
```
pffrandsen: This should be an interface - it is only used as such

GuutBoy:
@Rhaea Would it be possible for you to fix the few conflicts here so we can get this merged?

Rhaea:
Interleaving would theoretically break security as the invariant is that the tree from an output gate is MAC checked before output is delivered. If we now included the check in the next batch, this would mean that output is delivered potentially before the mac check is completed. Whether this happens or not depends on the evaluation strategy. 
But this does not mean there is not a more efficient solution. So sure, we can look at that at some later point yes.

PR #160, opened 2017-09-22
113 log performance

### LINKED ISSUE #113: Record/Log performance metrics
Record and log performance metrics, to help benchmark and optimize the framework and implemented protocols. Currently only limited metrics are being recorded and mainly when using the sequential evaluation strategy.
 
Useful metrics could include

- Timing of protocol execution
- Number of executions of key protocols (multiplication, comparison, etc.)
- Total communication amount
- Communication pr. protocol batch/round (e.g., max/min/mean/avg)
- Number of communication round/batches
- Number of native protocols pr batch (e.g., max/min/mean/avg)
- Perhaps some protocol suite specific metrics such as spdz MacChecks?

Rhaea:
Nearly, and could be argued to fix #113. Besides introducing a performance logger, this PR removes the createResourcePool method from the ProtocolSuite interface and moves this responsibility to the individual tests (or abstract test classes in this case). The test framework still makes sure the network is connected and disconnects when the test is done. 
The performance logger is able to log the following things:
- Running times on applications. Also lets the user know which evaluation strat and protocol suite was used to run the app. 
- Network activity which includes:
  - Number of times we received data
  - How much was received in total from each party
  - The min, max and avg amount of bytes received per batch
- Native protocol counter which includes the total amount evaluated as well as the min, max and avg per batch. 

When logging is enabled, decorators will be applied to the evaluators, the SCE and the network. Unfortunately for the network part, we can only log what is the actual message lengths send and received. If the messages are wrapped in a stupid way when send, we will not know with this generic logging tool. Still useful though. 

What is still missing: Logging of 'important protocols' - e.g. comparison. This type of logging requires deeper changes and are therefore not included yet. It is hard to really log all that might be useful. Often it is very specific things like 'how often is this class created' or similar which is not loggable by a generic framework. For now, we have the most important things logged.


pffrandsen:
I have myself implemented a simple count logging of the BuilderFactoryNumeric - it is contrary to your comment quite easy and very useful. I would expect something similar within this branch

core/src/main/java/dk/alexandra/fresco/suite/ProtocolSuite.java
```diff
@@ -67,34 +50,32 @@ ResourcePoolT createResourcePool(
...
    */
-  class DummyRoundSynchronization<ResourcePoolT extends ResourcePool> implements
-      RoundSynchronization<ResourcePoolT> {
+  class DummyRoundSynchronization<ResourcePoolT extends ResourcePool>
+      implements RoundSynchronization<ResourcePoolT> {
 
     @Override
-    public void finishedBatch(
-        int gatesEvaluated, ResourcePoolT resourcePool, SCENetwork sceNetwork) {
+    public void finishedBatch(int gatesEvaluated, ResourcePoolT resourcePool,
```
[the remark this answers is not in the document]
Rhaea: It is google code style through eclipse's formatter. If it split into two lines, it is because the first line was too long.

core/src/main/java/dk/alexandra/fresco/framework/PerformanceLogger.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  private void reset() {
+    this.networkLogger.clear();
+    this.runtimeLogger.clear();
+  }
+
+  private void newline(String msg, StringBuilder sb) {
+    sb.append(msg);
+    sb.append(System.lineSeparator());
```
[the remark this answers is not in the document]
Rhaea: Really? How would you pretty-print anything then?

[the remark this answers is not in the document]
pffrandsen: You are logging which gives a standardized logformat. This format is circumvented by newlines. So "Never use newline in a logger statement."

If you want it pretty (by your standards) then do not log it. Write it directly to a file or System.out.
I would prefer text that can be grepped for and searched through. Pretty for me means parseable, not human readable.

core/src/main/java/dk/alexandra/fresco/suite/ProtocolSuite.java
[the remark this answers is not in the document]
pffrandsen: Well. The previous format was designed by a human. And on two lines, you are changing that format - apparently due to an aggressive setting. My similar version of checkstyle does not need to rearrange the code. I would encourage you to refrain from doing these rearrangement of the code unless they are needed by checkstyle..

core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedProtocolEvaluatorPerformanceDecorator.java
```diff
@@ -0,0 +1,31 @@
...
+import dk.alexandra.fresco.framework.network.SCENetworkSupplier;
+import dk.alexandra.fresco.framework.sce.resources.ResourcePool;
+import java.io.IOException;
+
+public class BatchedProtocolEvaluatorPerformanceDecorator<ResourcePoolT extends ResourcePool, Builder extends ProtocolBuilder>
+    extends BatchedSequentialEvaluator<ResourcePoolT, Builder> {
+
+  private PerformanceLogger pl;
+
+  public BatchedProtocolEvaluatorPerformanceDecorator(PerformanceLogger pl) {
```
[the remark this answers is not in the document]
Rhaea: So just to be clear: you're saying make the evaluator decorator take an evaluator as a constructor parameter? If this is the case, it doesn't quite work since processBatch is called within the evaluator itself, thus avoiding any decorater. If this is not the case, please state what you intent (or drop by and explain :) )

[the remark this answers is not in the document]
pffrandsen: Currently there exists two different designs for evaluating batches. 
SequentialEvaluator has it's own internal
BatchedSequentialEvaluator uses the BatchStrategy
Introduce an interface that evaluates a batch of protocols and use it both places with a different implementation. Add a third which is decorating performance logging

core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedStrategy.java
```diff
@@ -116,4 +108,4 @@
       sceNetworkSupplier.nextRound();
     }
   }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
Rhaea: Google code style - I have (now had, since this apparently causes too many issues) a trigger within eclipse that formatted the file upon save. This is why some files are touched, but not changed beside formatting.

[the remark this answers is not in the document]
pffrandsen: Well, it can hardly be google style to blame.

I would expect a cleanup of this PR to remove the clutter.

Rhaea:
True - this you can do. I was so focused on logging during evaluation that I did not think about the construction phase. I will add this as well.

core/src/main/java/dk/alexandra/fresco/framework/network/NetworkPerformanceDelegate.java
```diff
@@ -0,0 +1,91 @@
...
+import dk.alexandra.fresco.framework.PerformanceLogger;
+import dk.alexandra.fresco.framework.configuration.NetworkConfiguration;
+import dk.alexandra.fresco.framework.util.Pair;
+import java.io.IOException;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class NetworkPerformanceDelegate extends PerformanceLogger implements Network {
+
+  private Network network;
```
pffrandsen: My previous comment was apparently confusing.

The class should be called NetworkLoggingDecorator. 
The inner field decorated or delegate.

The inheritance is currently used for passing the partyid into the logging. Change the PerformanceLogger to an interface and let print tak the party as an argument? Or a decorated logger.

Rhaea: Yea okay - I was indeed confused :) - will do

core/src/main/java/dk/alexandra/fresco/framework/PerformanceLogger.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Class for getting performance numbers. Use the EnumSet to indicate which parameters to log. Note
+ * that network logging is done on ONLY the data received and used within FRESCO. This means that if
+ * the network implementation uses double the bytes to wrap the messages, this will not show.
+ * 
+ * @author Kasper Damgaard
+ *
+ */
+public abstract class PerformanceLogger {
```
pffrandsen: From previous comment - unaddressed:
The inheritance is currently used for passing the partyid into the logging. Change the PerformanceLogger to an interface and let print take the party as an argument? Or a decorated logger. Inheritance is overkill here

core/src/main/java/dk/alexandra/fresco/framework/sce/SCELoggingDecorator.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Override
+  public <OutputT> OutputT runApplication(Application<OutputT, Builder> application,
+      ResourcePoolT resources) {
+    long then = System.currentTimeMillis();
+    OutputT res = this.delegate.runApplication(application, resources);
+    long now = System.currentTimeMillis();
+    long timeSpend = now - then;
+      this.runtimeLogger.add(new RuntimeInfo(application, timeSpend, protocolSuiteName));
```
pffrandsen: curious tabbing, does checkstyle approve?

core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedStrategy.java
```diff
@@ -58,34 +53,24 @@
...
+  @Override
+  public <sceNetwork extends SCENetwork & SCENetworkSupplier> void processBatch(
+      ProtocolCollection<ResourcePoolT> protocols, ResourcePoolT resourcePool, sceNetwork sceNetwork)
+          throws IOException {
+    Network network = resourcePool.getNetwork();
     int round = 0;
-
     while (protocols.size() > 0) {
-      evaluateCurrentRound(protocols, sceNetwork, channel, rp, network, round);
+      evaluateCurrentRound(protocols, sceNetwork, 0, resourcePool, network, round);
```
pffrandsen: I agree - channel is obsolete - why not remove it from the interface?

core/src/main/java/dk/alexandra/fresco/framework/sce/resources/ResourcePool.java
```diff
@@ -3,26 +3,23 @@
  *
  * This file is part of the FRESCO project.
  *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- * copies of the Software, and to permit persons to whom the Software is
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
```
pffrandsen: Reiterating: If no changes to this file, then revert to the original.

core/src/main/java/dk/alexandra/fresco/framework/sce/resources/ResourcePoolImpl.java
```diff
@@ -3,26 +3,23 @@
  *
  * This file is part of the FRESCO project.
  *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
- * of this software and associated documentation files (the "Software"), to deal
- * in the Software without restriction, including without limitation the rights
- * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
- * copies of the Software, and to permit persons to whom the Software is
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
```
pffrandsen: Reiterating: If no changes to this file, then revert to the original.

core/src/main/java/dk/alexandra/fresco/suite/ProtocolSuite.java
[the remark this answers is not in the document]
pffrandsen: This is still a less preferred alignment for me

core/src/test/java/dk/alexandra/fresco/framework/builder/BuildStepTests.java
```diff
@@ -47,17 +46,15 @@ public void test() throws Exception {
...
                 // iterate
-                (state) -> state.round < numIterations,
-                (seq, state) -> {
-                  List<Integer> roundsSoFar = state.rounds;
-                  roundsSoFar.add(state.round);
-                  return new IterationState(state.round + 1, roundsSoFar);
-                }).seq((seq, state) -> {
-                  return () -> state.rounds;
-                });
+                (state) -> state.round < numIterations, (seq, state) -> {
```
pffrandsen: This seems slightly confusing - I preferred the old line structure

core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedStrategy.java
Rhaea: Mostly because we would like in theory to play around with the idea of multiple channels - this enables threading such that a batch can be split into multiple worker threads - each with their own channel. 
This quickly becomes complicated, and was the reason we pulled it from the framework. Still - it might come in handy at some point

Rhaea:
I now introduced logging of basic protocols as well as comparisons - both for numeric. Let me know if you agree with this style of doing it. The point is that the dummy suite(s) will never be used in production anyway, so we might as well use logging within them. I could find no clever way of printing the logs, so therefore I used a public static final map. Not pretty, but works. Other ideas are most welcome.

core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedStrategy.java
pffrandsen: Ideas are free, code is not.

I would suggest to keep the idea but remove these legacy elements from the actual code making the future expansion possible without regard to the old version.

There are several ways of expanding the throughput over channels, the now removed implementation was not very good.

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/ProtocolBuilderNumeric.java
```diff
@@ -10,10 +10,10 @@
 public class ProtocolBuilderNumeric extends
     ProtocolBuilderImpl<ProtocolBuilderNumeric> {
 
-  private final BuilderFactoryNumeric factory;
+  protected final BuilderFactoryNumeric factory;
   private BasicNumericContext basicNumericContext;
-  private Numeric numeric;
-  private Comparison comparison;
+  protected Numeric numeric;
+  protected Comparison comparison;
```
pffrandsen: These three does not need to be protected? Make them private.

Issue #120, opened 2017-07-07
Add Javadoc


GuutBoy:
Add javadoc documentation. The goal would be to have every public interface, class and method should be documented. Central classes having higher priority.

PR #161, opened 2017-10-03
Added JavaDoc to almost all classes and interfaces in core.

Rhaea:
Also looked a bit at spdz. Removed ignored tests and created a new one for the static spdz data supplier. Goes a long way towards completing #120


pffrandsen:
Please fix the (quite immense) spelling error. It is not a "computation directive", rather a "computation directory"

Also correct the javadoc of Numeric
The interface should not read: 
@return A protocol computing a <op> b

but rather something like
@return A delayed result computing a <op> b

Protocols are at this point implementation details - a reasonable implementation could just return lambdas some of the places.

core/src/main/java/dk/alexandra/fresco/framework/DRes.java
```diff
@@ -1,9 +1,9 @@
 package dk.alexandra.fresco.framework;
 
 /**
- * Is the root of all calculations in fresco.
+ * Is the root of all calculations in FRESCO.
  * This is intended to open up for some protocols to be implemented by producers.
- * Both NativeProtocol and ProtocolProducer can be a computation
+ * Both {@link NativeProtocol} and intermediate results can be a DRes.
  *
```
GuutBoy: This is quite central class. I think the documentation here really needs improvement. For example, you may want to mentioned that DRes is short for "deffered result" and that is what this class is exactly supposed to model. I think the documentation still has some reference to "computation" as well.

Rhaea: True indeed - I actually did not know what the abbreviation was. Now I can do this in more detail.

Rhaea:
@GuutBoy About the code coverage: This is almost impossible to fix since the changes are so slight to the code (almost exclusively javadoc changes). I would have to do something artificial in order to meet the requirement. So I request that this time, we merge even if the code coverage is angry. 
Also, for some reason, the checks will not be processed anymore, so we cannot even be sure they pass.

pffrandsen:
You addressed 50 % of my comment, the rest is:
Correct the javadoc of Numeric (and other builders)
The interface should not read:
@return A protocol computing a b

but rather something like
@return A deferred result computing a b

Protocols are at this point implementation details - a reasonable implementation could just return lambdas some of the places.

core/src/main/java/dk/alexandra/fresco/lib/crypto/mimc/MiMCEncryption.java
```diff
@@ -34,6 +8,14 @@
 import dk.alexandra.fresco.lib.field.integer.BasicNumericContext;
 import java.math.BigInteger;
 
+/**
+ * MiMC encryption is a symmetric cipher which can be used to great effect when utilizing it within
+ * MPC. Implementation is done using <a href="https://eprint.iacr.org/2016/492.pdf">this paper</a>
+ * as base line. Encryption is both simple and fast, but decryption is a heavy operation. 
+ * 
+ * @author Kasper Damgaard
```
pffrandsen: There are still a few at author left

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/AdvancedNumeric.java
```diff
@@ -109,47 +158,56 @@
...
+   * 
+   * @param input The number to know the bit length of
+   * @param maxBitLength The maximum bit length this number can have (if unknown, set this to the
+   *        modulus bit size)
+   * @return A deferred result computing the bit length of the input number.
+   */
   DRes<SInt> bitLength(DRes<SInt> input, int maxBitLength);
 
+  /**
+   * Compute the inverse of x within the field of operation
```
pffrandsen: I am so sorry, but checkstyle requires a period in the first sentence of a javadoc.
"First sentence of Javadoc is incomplete (period is missing) or not present."
You should make sure that this branch does not have javadoc violations - at lest on the changed/updated files

GuutBoy:
@Rhaea If you only did javadoc changes the code coverage for the patch should be 100%. You must have done some code changes as well. Is it really not possible to cover those parts?

Rhaea:
@GuutBoy I guess, but I changed >300 files, and locating the few where I changed actual code is kinda time consuming and not well spent imo. Also, I think that maybe me changing alignment in a file would trigger the code coverage problem. I do not recall implementing any new functionality at all in this PR.

GuutBoy:
Ok, I will let it pass this once, and merge before this starts conflicting too much.

PR #166, opened 2017-11-13
Test coverage

MichaelStausholm:
Some minor changes + tests to improve test coverage.

There might be some old tests that are no longer needed / test already tested code.


core/src/main/java/dk/alexandra/fresco/lib/math/bool/log/Logarithm.java
```diff
@@ -80,9 +80,6 @@ public Logarithm(List<DRes<SBool>> number) {
    * Computes the floor(log_2(x)).
    */
   private int log2(int n) {
-    if (n <= 0) {
```
GuutBoy: Why delete this check?

core/src/test/java/dk/alexandra/fresco/framework/util/TestDetermSecureRandom.java
```diff
@@ -1,11 +1,11 @@
 package dk.alexandra.fresco.framework.util;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertTrue;
 
+import org.junit.Test;
 import dk.alexandra.fresco.framework.MPCException;
+import static org.junit.Assert.assertArrayEquals;
```
GuutBoy: Static imports should be at the top

core/src/main/java/dk/alexandra/fresco/lib/math/bool/log/Logarithm.java
MichaelStausholm: It's a private method and could never be invoked with n<=0

core/src/test/java/dk/alexandra/fresco/framework/util/TestDetermSecureRandom.java
MichaelStausholm: Fixed.

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMerge.java
```diff
@@ -66,7 +66,7 @@ private void merge(int first, int length, int step, ProtocolBuilderBinary builde
         }
         return null;
       });
-    } else if (length == 2) {
```
Rhaea: I would say either remove it entirely if it shouldn't be there, or have it stay. No point in having old incorrect code lying around in comments.

core/src/main/java/dk/alexandra/fresco/lib/compare/MiscBigIntegerGenerators.java
```diff
@@ -90,9 +90,9 @@ public MiscBigIntegerGenerators(BigInteger modulus) {
 
     for (int i = 1; i <= l; i++) {
       int k = i;
-      if (i >= m) {
```
Rhaea: Again, if this is the case, just remove it. Can keep the actual comment though as that might be useful.

core/src/main/java/dk/alexandra/fresco/lib/helper/bristol/BristolCircuitParser.java
```diff
@@ -111,8 +107,7 @@ private void parseLine(String line, ProtocolBuilderBinary builder) throws IOExce
     // System.out.println("IN: " + Arrays.toString(in));
     // System.out.println("OUT: " + Arrays.toString(out));
 
-    switch (type) {
```
Rhaea: Switch stmt makes it harder to do 100% testing? I see no other reason for changing this.

core/src/test/java/dk/alexandra/fresco/framework/util/TestBinaryMatrix.java
```diff
@@ -6,10 +6,14 @@
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
+
 import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.List;
 import java.util.Random;
+
```
Rhaea: I think the code style checker complains about this space :)

core/src/test/java/dk/alexandra/fresco/lib/compare/bool/eq/TestBinaryEquality.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import dk.alexandra.fresco.framework.value.SBool;
+import dk.alexandra.fresco.framework.DRes;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestBinaryEquality {
```
Rhaea: Class name could probably be a bit better. What about TestBinaryEQInputLength? The other name indicates a different set of tests.

core/src/test/java/dk/alexandra/fresco/lib/math/bool/add/TestFullAdder.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import dk.alexandra.fresco.framework.value.SBool;
+import dk.alexandra.fresco.framework.DRes;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class TestFullAdder {
```
Rhaea: Same for this name. Not sure what the best approach is, but a longer class name is needed I guess

core/src/main/java/dk/alexandra/fresco/lib/collections/sort/OddEvenMerge.java
GuutBoy: Agree, please dont leave code commented out. Either keep it or delete it. This is an issue a few other places

core/src/main/java/dk/alexandra/fresco/lib/helper/bristol/BristolCircuitParser.java
GuutBoy: I don't really see the reason either. At least it should be `if ... else ...` i guess. Also, please delete all the commented code.

GuutBoy: ... also you could reduce the complexity of this code by having just a single return statement at the end. There is really no reason to have them in each case.

core/src/test/java/dk/alexandra/fresco/framework/util/TestDetermSecureRandom.java
GuutBoy: Sorry, but static and non static should be in separate blocks and each block should be ordered alphabetically. https://google.github.io/styleguide/javaguide.html#s3.3-import-statements . It's really easier to do this organization automatically. Let me know if you need help setting up the IDE.

core/src/test/java/dk/alexandra/fresco/lib/compare/bool/eq/TestBinaryEquality.java
MichaelStausholm: I don't agree. I believe the class name should reflect the class being tested and then the method names can indicate exactly what is being tested. This is the naming model we've used for the other tests as well, eg. "TestBitSetUtils", "TestCreditRater", etc.

core/src/main/java/dk/alexandra/fresco/lib/helper/bristol/BristolCircuitParser.java
MichaelStausholm: The switch is hard to test (apparently it leaves out 3 branches somehow), but I have added else statements.

core/src/test/java/dk/alexandra/fresco/lib/compare/bool/eq/TestBinaryEquality.java
GuutBoy: I agree with Michael. This way you can extend the test class with more tests on the same class without having to change the name.

Rhaea: Okay, makes sense. Ignore comment then.

PR #167, opened 2017-11-13
Merged SCENetwork and SCENetworkSupplier

Rhaea:
I was thinking that this merge at least makes some sense. Merging Network and SCENetwork is still not the greatest idea if you ask me. That would create overloaded send and receive methods and be confusing to developers. 
Let me know what you think.


Rhaea:
No problem. Only took 15 min to do, and was to start the discussion on something concrete. I would love to see a solution which prevents protocol developers from using the incorrect send and receive methods. I will close this PR, and look forward to seeing your concept in code.

Issue #177, opened 2017-11-22
Test KryoNetwork


GuutBoy:
Various exceptions thrown are not tested. 
Private class Registrator is not tested, probably this class should just be a private method. 

Also encrypting traffic using AES is not tested, these parts should simply be deleted. We will handle secure channels at a later stage.

Issue #174, opened 2017-11-22
Delete the AES class


GuutBoy:
It is completely not tested and never used. It would be preferable if we just use something standard for secure channels like TLS. We will look into this in a later release.

PR #185, opened 2017-11-24
Removed AES fixing #174 and fully tested the KryoNetNetwork fixing #177

Rhaea:
Quick fix for the two issues #174 and #177 
One of the tests will take around 17 seconds no matter what since we need to wait for timeout.


core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
```diff
@@ -1,28 +1,29 @@
...
+import com.esotericsoftware.kryonet.Client;
+import com.esotericsoftware.kryonet.Connection;
+import com.esotericsoftware.kryonet.EndPoint;
+import com.esotericsoftware.kryonet.Listener;
+import com.esotericsoftware.kryonet.Server;
+import com.esotericsoftware.minlog.Log;
+
+import dk.alexandra.fresco.framework.MPCException;
+import dk.alexandra.fresco.framework.configuration.NetworkConfiguration;
+
```
GuutBoy: Wrong sort order of imports

Issue #182, opened 2017-11-22
Test LP package


GuutBoy:
Multiple classes in the LP package missing tests of dimension checking

PR #186, opened 2017-11-27
Lp tests #182

GuutBoy:
Covered the LP package with tests.

- Removed some dimension checking in classes that should only be used by the LPSolver (which already checks)
- Revived some old tests for exception handling
- Tested the debugging print out for the LPSolver and LPTableau, including changing how the debug flag is set in LPSolver. I am not sure if the solution is sort of a hack, but it should force you to override LPSolver to get the debug info
- Added tests for both pivot rules.


core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -90,12 +91,12 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
...
       }
 
-      LPState initialState = new LPState(BigInteger.ZERO, tableau, updateMatrix, null, pivot,
+      LpState initialState = new LpState(BigInteger.ZERO, tableau, updateMatrix, null, pivot,
           enumeratedVariables, initialBasis, pivot);
       return () -> initialState;
     }).whileLoop(state -> !state.terminated(), (seq, state) -> {
       iterations++;
-      if (debugLog) {
+      if (getDebuggable()) {
```
pffrandsen: In general I see the term "isDebug()" rather than this - could we rename it?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPTableau.java
```diff
@@ -45,10 +45,16 @@ public LPTableau(Matrix<DRes<SInt>> C, ArrayList<DRes<SInt>> B,
...
+   * Opens and outputs the plaintext values of this tableau for debugging.
+   * 
+   * @param builder a builder to provide the open and print functionality
+   * @param ps a PrintStream on which to print the debugging information
+   */
+  public void debugInfo(ProtocolBuilderNumeric builder, PrintStream ps) {
+    builder.debug().openAndPrint("C: ", C, ps);
+    builder.debug().openAndPrint("B: ", B, ps);
+    builder.debug().openAndPrint("F: ", F, ps);
+    builder.debug().openAndPrint("z: ", z, ps);
```
pffrandsen: At this point it would be nice to be able to reconfigure the output, so it is usable in the test - the test should inspect the debugged variables. However this is not possible with the current implementation of debug

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
```diff
@@ -0,0 +1,602 @@
...
+      // if (diff.compareTo(BigInteger.ZERO) > 0) {
+      // minNominator = updatedB[i];
+      // minDenominator = updatedColumn[i];
+      // exitingIndex = i;
+      // }
+      // }
+      // }
+      // }
+      // return exitingIndex;
+      return 0;
```
pffrandsen: Kill dead code

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
```diff
@@ -0,0 +1,602 @@
...
+            }
+          };
+          runApplication(app);
+          String output = bytes.toString("UTF-8");
+          System.out.println(output);
+          assertTrue(output
+              .contains("C: \n" + "0, 0, 0, \n" + "1, 1, 1, \n" + "2, 2, 2, \n" + "3, 3, 3,"));
+          assertTrue(output.contains("B: \n" + "10, 10, 10, 10,"));
+          assertTrue(output.contains("F: \n" + "0, 2, 3,"));
+          assertTrue(output.contains("z: \n" + "0"));
```
pffrandsen: Nice test, maybe recode it so future work on the logging only needs to change a single line

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
```diff
@@ -0,0 +1,602 @@
...
+            s.replaceAll("\r", "");
+            assertTrue(s.contains("C:"));
+            assertTrue(s.contains(
+                "C: \n" + "1, 0, 0, 1, 0, 0, \n" + "0, 1, 0, 0, 1, 0, \n" + "0, 0, 1, 0, 0, 1,"));
+            assertTrue(s.contains("B: \n" + "1, 2, 3, "));
+            assertTrue(s.contains("F: \n" + "-1, -1, -1, 0, 0, 0, "));
+            assertTrue(s.contains("z: \n" + "0"));
+            assertTrue(s.contains("Basis [1]: \n" + "4, 5, 6,"));
+            assertTrue(s.contains("Basis [4]: \n" + "1, 2, 3,"));
+            assertTrue(s.contains("Update Matrix [1]: \n" + "1, 0, 0, 0, \n" + "0, 1, 0, 0, \n"
```
pffrandsen: Ditto - consider recoding tests on output stream so they have a input, key and an expected value

core/src/main/java/dk/alexandra/fresco/lib/lp/LPTableau.java
GuutBoy: I am not exactly sure what you are thinking here.

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
GuutBoy: Not sure what you are thinking here. Just put everything in one assertTrue?

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
GuutBoy: Done

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
GuutBoy: Done

core/src/main/java/dk/alexandra/fresco/lib/lp/LPTableau.java
pffrandsen: Basically your tests are working from the design that the performance logger outputs strings. This entails testing on the form of String.contains("KeyOfMyIndex") and String.contains(String.valueOf(myKeyValue)) 

A more reasonable design in this context exposes the performance counters more directly. My suggestion is in this branch to alter the tests such that they introduce a helper method take the performance result (the string), a key ("z") and a value (z) and tests it currently via the string, but is easily reconfigured should the logger interface receive a bit of lova and change

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
GuutBoy: As above.

Issue #181, opened 2017-11-22
Test SecureComputationEngineImpl


GuutBoy:
Timeout/interrupt exceptions are never tested.

Issue #180, opened 2017-11-22
Test ResourcePoolImpl


GuutBoy:
getSecureRandom never tested, likely because the Dummy suites never uses this.

Issue #179, opened 2017-11-22
Test InMemoryStorage


GuutBoy:
Some simple conditionals are never tested

Issue #178, opened 2017-11-22
Test FilebasedStreamedStorageImpl


GuutBoy:
Exception handling not tested

PR #194, opened 2017-11-28
Sce tests

GuutBoy:
Various tests to cover `sce` package:

- Test of SecureComputationEngineImpl fixing #181 
- Test of ResourcePoolImpl fixing #180 
- Test of InMemoryStorage fixing #179 
- Test of FileBasedStreamedStorage fixing #178


core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
```diff
@@ -16,79 +16,99 @@
...
+   *
+   * <p>
+   * Defaults to a value of 10 minutes.
+   * </p>
+   *
+   * @param value value of the limit
+   * @param unit the time unit of the limit
+   */
+  public void setRunTimeout(int value, TimeUnit unit) {
+    timeOutValue = TimeUnit.NANOSECONDS.convert(value, unit);
```
pffrandsen: Use Duration

GuutBoy: Done.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
```diff
@@ -16,79 +17,98 @@
...
+  /**
+   * Sets the time out duration for waiting for output from running applications.
+   *
+   * <p>
+   * Defaults to a value of 10 minutes.
+   * </p>
+   *
+   * @param timeOut the new time out duration
+   */
+  public void setRunTimeout(Duration timeOut) {
```
pffrandsen: Exposing this as a setter seems wrong.

It belongs in the construction - and is currently only exposed in order to do tests

core/src/main/java/dk/alexandra/fresco/framework/sce/resources/storage/FilebasedStreamedStorageImpl.java
```diff
@@ -27,85 +26,72 @@
...
+   * Creates an instance of the file based streamed storage. For non-streamable object, the given
+   * internal storage is used.
+   *
+   * @param internalStorage The storage used for non-streamable objects
+   * @param inputs a map from store names to input streams reading from the named store
+   * @param inputs a map from store names to output streams writing to the named store
+   */
+  protected FilebasedStreamedStorageImpl(Storage internalStorage,
+      ConcurrentHashMap<String, ObjectInputStream> inputs,
+      ConcurrentHashMap<String, ObjectOutputStream> outputs) {
```
pffrandsen: Why concurrent - that seems steep?

GuutBoy: We used to have multiple threads access these maps via the FiledBasedStreamedStorage. With the current evaluators, I suppose they no longer need to be concurrent.

pffrandsen: I agree, however synchronous access to a storage also implies a Serialization transaction level on the storage, which in turn renders the synchronous bit irrelevant. My best estimate is to make the resource pool and in turn the storage "splittable"

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
GuutBoy: Previously the timeout arbitrarily set to 10 minutes and was unchangeable, which definitely seemed wrong. I dont I have strong opinion about whether it should be configured in the constructor or as a setter, I just think it should be possible for the user to define. Is it not desirable that one could change the timeout during the lifetime of a SCE object? If not, I guess the constructor is the way to go.

pffrandsen: Well

Either use start or run application, 
If you want control, use start - if just default behaviour, then run is fine.

My point is start is for the advanced usages - and no, you never need to change the timeout after the creation.

core/src/main/java/dk/alexandra/fresco/framework/sce/resources/storage/FilebasedStreamedStorageImpl.java
GuutBoy: So, for now I am thinking, just simple Map's. In the longer run the whole Storage package probably needs a workover.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
GuutBoy: How about passing it as a parameter to the runApplication method where it is used?

pffrandsen: Good idea, but leave a default implementation in the interface of SecureComputationEngine in order to keep a version of run and start similar

GuutBoy: How about this way then?

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
```diff
@@ -4,55 +4,78 @@
...
+      ResourcePoolT resources, Network network) {
+    return runApplication(application, resources, network, DEFAULT_TIMEOUT);
+  }
+
+  /**
+   * Executes an application based on the current configuration.
+   *
+   * @param application the application to evaluate
+   * @param resources the resource pool used in evaluation
+   * @param network the network
```
Rhaea: Same here

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
```diff
@@ -4,55 +4,78 @@
...
-   * In a normal application this should be the normal way to start an application since there
-   * need to be allocated resources (the resource pool) and allowed for parallel work.
+   * is not setup before (e.g. connected to other parties etc.), the SecureComputationEngine will do
+   * the setup phase before running the application. <br/>
+   * In a normal application this should be the normal way to start an application since there need
+   * to be allocated resources (the resource pool) and allowed for parallel work.
    *
    * @param application The application to evaluate.
+   * @param resources the resource pool used in evaluation
    * @param network the network to run the computations on
```
Rhaea: And same here ofc.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
```diff
@@ -4,55 +4,78 @@
...
    * Executes an application based on the current configuration.
+   * <p>
+   * By default this method times out after waiting 10 minutes for output.
+   * </p>
    *
-   * @param application The application to evaluate.
-   * @param network the network to run the computations on
+   * @param application the application to evaluate
+   * @param resources the resource pool used in evaluation
+   * @param network the network
```
Rhaea: Maybe add that the network is expected to be connected before running this method.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
```diff
@@ -4,55 +4,78 @@
...
+   * Initializes the SecureComputationEngine. This method is idempotent - and sets up the proces
+   * queue and initializes the protocol suite.
    */
   void setup();
 
   /**
-   * Ensures that resources are shut down properly. Network is disconnected
-   * and sockets are released.
+   * Ensures that resources are shut down properly. Network is disconnected and sockets are
+   * released.
```
pffrandsen: Network is no longer the responsibility of the SCE

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
pffrandsen: Hmm. that actually goes for every parameter - they need to be in "working condition"
A bit superfluous I would say

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
GuutBoy: Ok, I will update the comment.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngine.java
Rhaea: Well, that may be correct. Since there is no "connect" method in the network interface anyways, there may not be much point in pointing it out.. Still though, I would add it. I try to expect users to be not that smart :)

GuutBoy: Yeah, I suppose you would get some sort of network related exception in the case of a non-connected network, which is as can be expected.

PR #200, opened 2017-12-05
Missing enum tests

MichaelStausholm:
Two trivial tests to help test coverage %


GuutBoy:
This does not seem to change coverage at all. I do not see the point?

PR #201, opened 2017-12-05
Logging framework tests

MichaelStausholm:
Some minor changes to the way logging decorators work and explicit tests for all the decorators.

The tests are not exactly elegant, so if anyone have an idea to make them nicer, let me know.


MichaelStausholm:
I guess one option would be to introduce a getLoggedValues() method returning a Map<String, Object> to the PerformanceLogger class. This could then produce the same key-value pairs that are actually printed in the log. This does however not make testing the of printToLog() methods any more stable.

core/src/main/java/dk/alexandra/fresco/framework/sce/SecureComputationEngineImpl.java
```diff
@@ -29,7 +29,7 @@
     implements SecureComputationEngine<ResourcePoolT, BuilderT> {
 
   private ProtocolEvaluator<ResourcePoolT, BuilderT> evaluator;
-  private ExecutorService executorService;
+  public ExecutorService executorService;
```
GuutBoy: This should be changed back to private.

core/src/test/java/dk/alexandra/fresco/logging/TestArithmeticLoggingDecorators.java
```diff
@@ -0,0 +1,171 @@
...
+    TestThreadRunner.run(f, conf);
+    
+    for(Integer pId: netConf.keySet()) {
+      List<PerformanceLogger> pl = DummyArithmeticBuilderFactory.performanceLoggers.get(pId);
+      
+      ListLogger testLogger = new ListLogger();
+      pl.get(0).printToLog(testLogger, pId);
+      String expectedOutput = pl.get(0).makeLogString(pId);
+      Map<String, Object> loggedValues = pl.get(0).getLoggedValues(pId);
+      Assert.assertThat(loggedValues.get(NumericLoggingDecorator.ARITHMETIC_BASIC_ADD), Is.is((long)5719386));
```
GuutBoy: Did you check that this is these are the correct numbers? I think we ought to test with computations generating more predictable numbers, say a single (few) additions, mults, etc.

core/src/test/java/dk/alexandra/fresco/logging/TestGenericLoggingDecorators.java
```diff
@@ -0,0 +1,301 @@
...
+    TestThreadRunner.run(f, conf);
+
+    for (Integer pId : pls.keySet()) {
+      PerformanceLogger performanceLogger = decoratedLoggers.get(0);
+      performanceLogger.printToLog(pls.get(pId), pId);
+      String expectedOutput = performanceLogger.makeLogString(pId);
+      Assert.assertThat(pls.get(pId).getData().get(0), Is.is(expectedOutput));
+
+      Map<String, Object> loggedValues = performanceLogger.getLoggedValues(pId);
+      Assert.assertThat(loggedValues.get(BatchEvaluationLoggingDecorator.BATCH_COUNTER), Is.is(3496361));
```
GuutBoy: Again, this number is not really obvious at all. This goes for many of these tests.

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -17,19 +19,44 @@
...
-  public abstract void printPerformanceLog(int myId);
+  public default void printPerformanceLog(int myId) {
+    printToLog(PerformanceLogger.log, myId);
+  }
+  
+  /**
+   * Prints any performance numbers picked up using the
+   * specified logger. 
+   */
+  public void printToLog(Logger log, int myId);
```
GuutBoy: I really dont like this construction. It complicates the interface and is only there for testing purposes.

core/src/test/java/dk/alexandra/fresco/logging/TestArithmeticLoggingDecorators.java
```diff
@@ -0,0 +1,171 @@
...
+import dk.alexandra.fresco.suite.dummy.arithmetic.DummyArithmeticBuilderFactory;
+import dk.alexandra.fresco.suite.dummy.arithmetic.DummyArithmeticProtocolSuite;
+import dk.alexandra.fresco.suite.dummy.arithmetic.DummyArithmeticResourcePool;
+import dk.alexandra.fresco.suite.dummy.arithmetic.DummyArithmeticResourcePoolImpl;
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import org.hamcrest.core.Is;
```
GuutBoy: Consider having these as static imports. It makes for more readable test code.

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -17,7 +19,7 @@
   public DecimalFormat df = new DecimalFormat("#.00");
   
   public enum Flag {
-    LOG_NETWORK, LOG_RUNTIME, LOG_NATIVE_BATCH;
+    LOG_NETWORK, LOG_EVALUATOR, LOG_NATIVE_BATCH;
```
Rhaea: Maybe we should consider removing this enum for good? It doesn't cover all loggers anyways, and no logger decorators are created by the framework itself anyways. You have to do this yourself. Vote to remove. It used to make sense when decorators was created within the framework.

GuutBoy: How was it used before?

MichaelStausholm: It is/was used to control if a certain decorator should be used when initializing the sce. It's only really used in the AbstractDummy*Tests

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -17,19 +19,38 @@
...
 
   /**
-   * Prints any performance numbers picked up.
+   * Prints any performance numbers picked up using the
+   * default logger.
    */
-  public abstract void printPerformanceLog(int myId);
+  public default void printPerformanceLog(int myId){
+    log.info(makeLogString(myId));
+  }
```
pffrandsen: One request: The performance logger should not print it. Give this method a different object (PerformancePrinter) and create a reasonable default in the core test framework (which is the only use currently)

core/src/test/java/dk/alexandra/fresco/logging/TestArithmeticLoggingDecorators.java
MichaelStausholm: I agree, but this is a general comment related to most tests. Should this be moved to a separate issue?

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
MichaelStausholm: I've made a suggestion implementation of a PerformancePrinter and refactored some code to it. Please check if it was something along those lines you had in mind.

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -12,24 +12,26 @@
...
+   * Produce a key-value map of logged values. The content
+   * of the returned map is context dependent. 
    */
-  public abstract void reset();    
+  public abstract Map<String, Object> getLoggedValues(int myId);
   
+  /**
+   * Make a generic log string from the logged values. 
+   */
+  public default String makeLogString(int myId){
```
pffrandsen: This also belongs in the performance printer

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -12,24 +12,26 @@
...
-  public abstract void printPerformanceLog(int myId);
+  public abstract void reset();    
   
   /**
-   * Resets any counters/maps/lists used. 
+   * Produce a key-value map of logged values. The content
+   * of the returned map is context dependent. 
    */
-  public abstract void reset();    
+  public abstract Map<String, Object> getLoggedValues(int myId);
```
pffrandsen: Given PerformanceLogger counts things the map could even be to integer?

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
```diff
@@ -12,24 +12,26 @@
...
-
-    public static final EnumSet<Flag> ALL_OPTS = EnumSet.allOf(Flag.class);
-  }
-
   /**
-   * Prints any performance numbers picked up.
+   * Resets any counters/maps/lists used. 
    */
-  public abstract void printPerformanceLog(int myId);
+  public abstract void reset();
```
pffrandsen: A public interface have by default public abstract methods - do not declare them.

core/src/main/java/dk/alexandra/fresco/logging/PerformanceLogger.java
MichaelStausholm: The NetworkLogger currently also logs bytes received pr. party, so the map should flexible enough to handle other datatypes, if we wish to support that level of logging.

pffrandsen: Just create a composite key with "BytesRecieved-<id>"

core/src/main/java/dk/alexandra/fresco/logging/DefaultPerformancePrinter.java
```diff
@@ -0,0 +1,16 @@
+package dk.alexandra.fresco.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultPerformancePrinter implements PerformancePrinter {
+
+  public Logger log = LoggerFactory.getLogger(PerformanceLogger.class);
```
GuutBoy: Probably this should be `public Logger log = LoggerFactory.getLogger(DefaultPerformancePrinter.class);`. Not `PerformanceLogger.class`.

core/src/main/java/dk/alexandra/fresco/logging/DefaultPerformancePrinter.java
```diff
@@ -0,0 +1,16 @@
...
+
```
GuutBoy: I agree we should add the `makeLogString`method to this class.

GuutBoy: I.e., not have it on the individual loggers.

core/src/main/java/dk/alexandra/fresco/logging/DefaultPerformancePrinter.java
```diff
@@ -0,0 +1,19 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultPerformancePrinter implements PerformancePrinter {
+
+  public Logger log = LoggerFactory.getLogger(PerformanceLogger.class);
+  
+  @Override
+  public void printPerformanceLog(PerformanceLogger logger, int myId) {
+    String s = "Logger for "+this.getClass().getName()+": ";
```
GuutBoy: There is a refactoring error here. It should probably be`logger.getClass().getName()`. An alternative Approach would be to remove this line and change the last line to `LoggerFactory.getLogger(logger.getClass()).info(s)`. The latter might give nicer output.

GuutBoy: Also, please run checkstyle.

PR #203, opened 2017-12-05
Minor tests and refactoring

MichaelStausholm:
Mixed bag of minor changes that should bring us to 100% coverage for core (combined with the other pull requests).

As far as I could see, the code in RepeatedRightShift could not be reached (atleast not if the class is accessed as it is supposed to be), but you might want to confirm.


core/src/test/java/dk/alexandra/fresco/lib/lp/TestLpBuildingBlocks.java
```diff
@@ -8,13 +8,22 @@
 import dk.alexandra.fresco.framework.DRes;
 import dk.alexandra.fresco.framework.value.SInt;
 import dk.alexandra.fresco.lib.collections.Matrix;
+import dk.alexandra.fresco.lib.lp.LPSolver.PivotRule;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import org.hamcrest.core.Is;
```
GuutBoy: These are already static imports.

Issue #205, opened 2017-12-08
Clean up MPCExceptions


n1v0lg:
Most MPCException instances are thrown when something non-malicious happens. These can just be Runtime exceptions and should be replaced by these. The MPCException instances thrown in the malicious case should be MaliciousExceptions.

pffrandsen:
Fixed on develop through the Mascot branch

PR #207, opened 2017-12-12
Actual exception handling in the SpdzResourcePool

pffrandsen:
This is a start for the exception branch, it is (almost) required for your mascot branch


GuutBoy:
Does this fix #205?

pffrandsen:
Not completely no. It presents a solution of the problem and fixes spdz resource pool creation

GuutBoy:
Ok, I will keep the issue open then.

PR #208, opened 2017-12-13
Remove static variable in performance log

### LINKED ISSUE #206: Bug in logging tests.
testNetworkLoggingDecorator sometimes fail with a nullpointer exception. Probably, the problems is that the network loggers are not collected in a concurrent way.

MichaelStausholm:
Some changes (and a bugfix which fixes #206 )for the logging framework.


core/src/test/java/dk/alexandra/fresco/suite/dummy/arithmetic/AbstractDummyArithmeticTest.java
```diff
@@ -94,21 +103,25 @@ protected void runTest(
                 KryoNetNetwork kryoNetwork = new KryoNetNetwork(partyNetConf);
                 if (logPerformance) {
                   NetworkLoggingDecorator network = new NetworkLoggingDecorator(kryoNetwork);
-                  pls.add(network);
+                  aggregate.add(network);
```
pffrandsen: This will not work as intented, the network has two values 
- minBytesReceived
- maxBytesReceived

Which will be summed.

PR #209, opened 2017-12-15
Kryonet refactor

Rhaea:
Nikolai recently discovered a bug in KryoNetNetwork which this PR should fix. Multiple test runs at least confirms this, but since it was a deadlock issue, one cannot say for certain. With high probability though, it's fixed :)

Also, if someone can explain why using free ports for the test testKryoNetConnectInterrupt results in an internal BindException I will be very happy to hear the reason. Happens only during a complete test run, and I see no mentions in the test output of the port it is trying to bind to.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -47,6 +47,16 @@
   <build>
     <plugins>
 
+	
+      <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-eclipse-plugin</artifactId>
+          <configuration>
+              <downloadSources>true</downloadSources>
+              <downloadJavadocs>true</downloadJavadocs>
+          </configuration>
+      </plugin>
+    
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -47,16 +47,6 @@
   <build>
     <plugins>
 
-	
-      <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-eclipse-plugin</artifactId>
-          <configuration>
-              <downloadSources>true</downloadSources>
-              <downloadJavadocs>true</downloadJavadocs>
-          </configuration>
-      </plugin>
-    
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
--- a/core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
+++ b/core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
@@ -6,6 +6,7 @@ import com.esotericsoftware.kryonet.Connection;
 import com.esotericsoftware.kryonet.EndPoint;
 import com.esotericsoftware.kryonet.Listener;
 import com.esotericsoftware.kryonet.Server;
+import com.esotericsoftware.kryonet.util.TcpIdleSender;
 import com.esotericsoftware.minlog.Log;
 import dk.alexandra.fresco.framework.configuration.NetworkConfiguration;
 import dk.alexandra.fresco.framework.util.ExceptionConverter;
@@ -14,8 +15,10 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentLinkedDeque;
@@ -181,6 +184,7 @@ public class KryoNetNetwork implements Network, Closeable {
             return;
           }
           // Multiple messages are needed to obtain the original large payload
+          logger.debug("Received data: " + data.length+". Expected before countdown: " + expected);
           ByteBuffer tmp = this.temporaryLists.get(fromPartyId);
           tmp.put(data);
           expected--;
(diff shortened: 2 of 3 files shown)


core/src/test/java/dk/alexandra/fresco/framework/network/TestKryoNetNetwork.java
```diff
@@ -189,22 +191,23 @@ public void run() {
...
     Thread t1 = new Thread(new Runnable() {
       
       @Override
       public void run() {
         NetworkConfiguration conf = new NetworkConfigurationImpl(1, parties);
         KryoNetNetwork network = null;
-        try{
+        try {
           network = new KryoNetNetwork(conf);
           Assert.fail("Should not be able to connect");
```
pffrandsen: Small suggestion - this could be a RuntimeException (for clarity), the Assert.fail throws a more uncontrolled exception

Rhaea:
The added test is indeed made for recreating the problem. The deadlock almost always occured when running the new test, so I thought that keeping it around would prevent the same issue from appearing without detection somewhat early on.

core/src/test/java/dk/alexandra/fresco/framework/network/TestKryoNetNetwork.java
Rhaea: Redid all asserts within threads since this did not actually cause the test to fail if they failed.

core/src/test/java/dk/alexandra/fresco/framework/network/TestKryoNetNetwork.java
```diff
@@ -1,226 +1,328 @@
...
+
+  @Before
+  public void reset() {
+    exceptions.clear();
+  }
+
+  /**
+   * Checks for exceptions and casts an Assert.fail if any was found
+   */
+  @After
```
pffrandsen: Taken from:
http://junit.org/junit4/javadoc/4.12/index.html?org/junit/After.html
"If you allocate external resources in a Before method you need to release them after the test runs. "

... not as a part of the test.
You should not fail in the after, it is intended for releasing external resources.

Rhaea:
@pffrandsen Yea this works just fine for some of the tests. However, some tests have assert statements in both threads. How do we check for those exceptions if we do not use the current style of adding exceptions to a list or similar and checking it afterwards? I removed the @After part, and if you insist on making it prettier, I hope you have time to do so yourself. The current solution should work and do what is required.

GuutBoy:
Attempted to simplify the tests using less threads and using the callable class to collect exceptions from threads.

core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
```diff
@@ -47,31 +46,23 @@
...
     this.conf = conf;
-    this.channelAmount = 1;
     this.clients = new HashMap<>();
     this.clientThreads = new ArrayList<>();
-    this.servers = new ArrayList<>();
-    this.queues = new ArrayList<>();
+    this.queues = new HashMap<>();
 
     // TODO: How to reason about the upper boundries of what can be send in a single round?
     int writeBufferSize = 1048576;
```
n1v0lg: I agree with the TODO note--we do need a way to reason about this/ configure it. Currently, the Mascot implementation relies heavily on batched messaging; attempting to generate a batch greater than 1000 triples causes a buffer overflow exception. Same goes for increasing the number of batches in the Fresco batched evaluation strategy. At the very least this parameter should be configurable but ideally the network implementation should automatically handle messages of arbitrary size.

Rhaea: I think I can fix this actually. Since we only send byte[]'s, it's just a matter of checking the size before sending, splitting up if needed. Will try to get this in today, but might be tomorrow. Also, @GuutBoy: Your attempt is nice, but Travis complains :) - can you fix this or should I do it?

n1v0lg: @Rhaea One thing I just noticed is that changing the buffer size has a significant impact on performance. In particular, if the buffer size is large, running on small messages (for instance for the networked component tests) is much slower than if the buffer size is small. This isn't surprising but nonetheless means that how we chunk the messages should still be configurable to allow fine-tuning the chunk size.

GuutBoy: @Rhaea I was not able to recreate the bug locally, and I really had to leave. If you can figure it out that would be great!

Also, I agree with @n1v0lg, the buffer size should really be configurable somehow, and would be great to be able to handle arbitrary size messages.

Rhaea: No problem. I will make it configurable as well. Good findings!
@GuutBoy: Same here with the bug. I will take a look at it today.

Rhaea:
Finally.. Jeez. Not sure if there are more race conditions left, and this was just a lucky run. Have not been able to recreate the bugs that Travis encounters at any point in time though, so it's hard to know. Should we commit a few non-commits to test before merging to develop?

core/pom.xml
```diff
@@ -47,6 +47,16 @@
...
 
+	
+      <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-eclipse-plugin</artifactId>
+          <configuration>
+              <downloadSources>true</downloadSources>
+              <downloadJavadocs>true</downloadJavadocs>
+          </configuration>
+      </plugin>
```
pffrandsen: Is this plugin needed or is it just for debugging?

core/src/main/java/dk/alexandra/fresco/framework/network/KryoNetNetwork.java
```diff
@@ -108,31 +125,31 @@ public ClientConnectThread(Client client, String hostname, int port, Semaphore s
...
     public void run() {
       boolean success = false;
       int maxRetries = 30;
+      int sleepTime = 200;
       int retries = 0;
       while (!success) {
         try {
           retries++;
-          client.connect(2000, hostname, port);
+          client.connect(1000, hostname, port);
```
pffrandsen: One second connection is very reasonable for unittests, but in real life it is very little time. Maybe defer it to a parameter?

core/pom.xml
Rhaea: Oh.. I forgot to remove that again. Although it definitely doesn't hurt to have it in there, it is not needed. Will remove again.

Rhaea:
Okay, 2 passes I am more comfortable with. 
I think we can make the network send async somewhat easily, but a simple local attempt didn't work. Let's just keep it as it is, and then we can finetune it in a later version

Issue #114, opened 2017-06-30
Adopt common code-style


GuutBoy:
We have adopted the google java code style. The old code should be moved to this style and the new style should be enforced in new pull requests.

GuutBoy:
I am closing this issue as it is a little too open ended. We should obviously keep focusing on consistent style.

Issue #106, opened 2017-06-30
Improve test coverage


GuutBoy:
The goal is 100% test coverage of the framework with prioritizing projects in the following order:
 - [x] Core 
 - [ ] Suite
 - [x] Demo
 - [x] Tools

Rhaea:
I had, and will have a few hours to work on improving the SPDZ test coverage. Just FYI. Will not put myself as assignee or change label for that. Should we? Not sure what the best course of action is.

GuutBoy:
I think it is a good idea to do, so we know what is going on.

Rhaea:
Started test coverage on demo within the branch https://github.com/aicis/fresco/tree/demos-test-coverage

GuutBoy:
I will close this to favor more specific issues.

PR #211, opened 2017-12-21
Spdz test coverage

MichaelStausholm:
Some refactoring and new testcases for the spdz suite (now on 100% coverage). 

These changes are part of issues #106 and #114.


suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/storage/DummyDataSupplierImpl.java
```diff
@@ -59,41 +59,36 @@ public DummyDataSupplierImpl(int myId, int numberOfPlayers) {
   public SpdzTriple getNextTriple() {
     if (numberOfPlayers == 2) {
       return triples_2party.get(myId);
-    } else if (numberOfPlayers == 3) {
+    } else {
       return triples_3party.get(myId);
     }
-    throw new RuntimeException("Not implemented yet");
```
Rhaea: I think it's nice to have an exception cast if someone tries to use the dummy supplier for more than 3 parties. As it stands, I guess an array out of bounds exception is cast if player >3 tries to call any method within this class.

MichaelStausholm: The RuntimeExceptions could never be reached. The constructor already does a check for numberOfParties > 3 (and casts a IllegalArgumentException).

Rhaea: Ah, nice. Well then, if tests go through, all is good by me.

demos/common/pom.xml
```diff
@@ -16,6 +16,12 @@
       <version>1.3</version>
     </dependency>
 
+	<dependency>
+      <groupId>dk.alexandra.fresco</groupId>
+      <artifactId>spdz</artifactId>
+      <version>${fresco-version}</version>
+      <type>test-jar</type>
+    </dependency>
```
pffrandsen: Why are we depending on the test jar in demos?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
```diff
@@ -7,37 +7,37 @@
  * An inputmask for player_i is random value r 
  * shared among parties so that only player_i knows the real value r.
  */
-public class SpdzInputMask implements Serializable{
+public class SpdzInputMask implements Serializable {
```
pffrandsen: Why is it serializable?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
```diff
@@ -7,37 +7,37 @@
...
-	 */
-	public BigInteger getRealValue(){
-		return realValue;
-	}
-
-	@Override
-	public String toString() {
-		return "SpdzInputMask [mask=" + mask + ", realValue=" + realValue + "]";
-	}
+
```
pffrandsen: Can be private

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
pffrandsen: There are actually a few classes that are Serializable - is that needed?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzKnownSIntProtocol.java
```diff
@@ -10,10 +10,10 @@
 public class SpdzKnownSIntProtocol extends SpdzNativeProtocol<SInt> {
 
   BigInteger value;
```
pffrandsen: This can be private?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -49,16 +55,21 @@ public SpdzMacCheckProtocol(SecureRandom rand, MessageDigest digest, SpdzStorage
...
 
-        pp = new SequentialProtocolProducer(
-            new SingleProtocolProducer<>(comm),
-            new SingleProtocolProducer<>(open));
+        pp = new SequentialProtocolProducer(new SingleProtocolProducer<>(comm),
+            new SingleProtocolProducer<>(openComm));
       } else if (round == 1) {
-        BigInteger alpha = storage.getSSK();
+        if (!comm.out()) {
+          throw new MPCException(
```
pffrandsen: This should be Malicious if it in fact is malicious

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -49,16 +55,21 @@ public SpdzMacCheckProtocol(SecureRandom rand, MessageDigest digest, SpdzStorage
...
+              "Malicious activity detected: Broadcast of commitments was not validated.");
+        }
+        if (!openComm.out()) {
+          throw new MPCException("Malicious activity detected: Opening commitments failed.");
```
pffrandsen: Ditto

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -69,42 +80,50 @@ public SpdzMacCheckProtocol(SecureRandom rand, MessageDigest digest, SpdzStorage
...
           a = a.add(aa.multiply(rs[index++])).mod(modulus);
         }
 
+        List<SpdzElement> closedValues = storage.getClosedValues();
         // compute gamma_i as the sum of all MAC's on the opened values times
         // r_j.
         if (closedValues.size() != t) {
-          throw new MPCException(
-              "Amount of closed values does not equal the amount of partially opened values. Aborting!");
+          throw new MPCException("Malicious activity detected: Amount of closed values does not "
```
pffrandsen: Ditto - maliciousException?

demos/common/pom.xml
MichaelStausholm: We moved some dummy/test classes from spdz/main into spdz/test (eg. dummy storage, fake tripple generation). These are however used for the demos.

Rhaea: We moved a few classes into the test part of the code since they had no place within main. In particular the DummyDataSupplier. Thus, we need the test jar to use this within the demos.

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
Rhaea: Yes, when using the file based streamed storage to pull the preprocessed data from disk, the storage requires objects to extend Serializable (to be used by the ObjectInputStream).

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
Rhaea: What do you mean? The exception? Isn't this only within the mascot branch currently?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
MichaelStausholm: The dk.alexandra.fresco.framework.sce.resources.storage.Storage class uses the Serializable interface to store objects. It can probably be changed, but requires some changes in core

demos/common/pom.xml
pffrandsen: Having thought this through, eliminating dummy data supplier from the spdz is the right way to go. However then you need to do a lot more work. Including building submaven projects for each supplier.

My suggestion is to move the dummy data suppler back into spdz. It is similar to the dummyprotocol that also is tolerated in the core

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/datatypes/SpdzInputMask.java
pffrandsen: Ok

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
pffrandsen: You are right, sorry

demos/common/pom.xml
Rhaea: I see your point - It would be easier to keep the dummy supplier within spdz. Also makes it possible to test applications without having to import the test jar.

PR #212, opened 2017-12-21
Removed fuelstation.

Rhaea:
As discussed with @GuutBoy it was not used, and too slow for any real application. Moved to private area for possible later modifications. We will have an easier time refactoring without having to keep unused code up to date.


pffrandsen:
Brilliant, this will really improve coverage.

PR #216, opened 2018-01-15
Async network

Rhaea:
After having spend so much time on the network problems, I thought to myself: There must be a simpler and better way. And it seems there is. My local testing at least, shows approx a 17% speed increase. This network also sends fewer packages than kryonet which should be beneficial overall.


Rhaea:
This PR should be modified depending on your comments to either remove KryoNet entirely or extract KryoNet as a separate tool (as Scapi is currently).

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+    SocketAddress sock = new InetSocketAddress(conf.getParty(conf.getMyId()).getPort());
+    try {
+      this.server = ServerSocketChannel.open();
+      this.server.bind(sock);
+      logger.info("Bound at " + sock);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to bind to " + sock, e);
+    }
+    for (int i = 1; i <= conf.noOfParties(); i++) {
+      this.queues.put(i, new ArrayBlockingQueue<>(1000000));
```
GuutBoy: It's not so nice to have this as a literal constant. I would prefer having it as a final static field, say `DEFAULT_QUEUE_SIZE`. Maybe this value should even be configurable? You could also use a `LinkedBlockingQueue` (I think that is the name) to avoid having to cap the size.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+      if (i != conf.getMyId()) {
+        Party p = conf.getParty(i);
+        SocketAddress addr = new InetSocketAddress(p.getHostname(), p.getPort());
+        try {
+          int retries;
+          int milliSecondsBetween;
+          if (timeout < 500) {
```
GuutBoy: Again, I am not so happy about these mysterious literal constants.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+    Thread t = new Thread(new Handshaker(this.server, partyIdToChannel, blocker, successes));
+    t.start();
+    for (SocketChannel channel : this.clients.values()) {
+      ByteBuffer id = ByteBuffer.allocate(1);
+      id.put((byte) this.conf.getMyId());
+      id.position(0);
+      ExceptionConverter.safe(() -> channel.write(id), "Not able to send my id");
+    }
+    ExceptionConverter.safe(() -> {
+      blocker.acquire();
```
GuutBoy: Wouldn't it be more natural use `t.join()` here? Maybe I am missing something.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+      return null;
+    } , "Blocker got interrupted");
+    boolean connected = true;
+    for (Boolean succ : successes) {
```
GuutBoy: Couldn't success be signaled from the handshake as a single `boolean`? I do not see the need for this list.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+      while (count < getNoOfParties() - 1) {
+        try {
+          SocketChannel channel = server.accept();
+          channel.configureBlocking(true);
+          ByteBuffer buf = ByteBuffer.allocate(1);
+          channel.read(buf);
+          buf.position(0);
+          int id = buf.get();
+          partyIdToChannel.put(id, channel);
+          successes.add(true);
```
GuutBoy: You do not really need this list. Just have a single `boolean` and set it to false if there is an exception. Or am I missing something?

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+   * be processed before closing the connection. In practice, the maximum amount of waiting time
+   * will never occur.
+   */
+  @Override
+  public void close() {
+    ExceptionConverter.safe(() -> {
+      this.receiverService.shutdownNow();
+      for (ExecutorService executorService : this.senderServices.values()) {
+        executorService.shutdown();
+        executorService.awaitTermination(500, TimeUnit.MILLISECONDS);
```
GuutBoy: Again with the literal constants.

core/src/test/java/dk/alexandra/fresco/suite/dummy/arithmetic/AbstractDummyArithmeticTest.java
```diff
@@ -100,7 +101,8 @@ protected void runTest(
               () -> new DummyArithmeticResourcePoolImpl(playerId,
                   noOfParties, drbg, mod),
               () -> {
-                KryoNetNetwork kryoNetwork = new KryoNetNetwork(partyNetConf);
+                //Network kryoNetwork = new KryoNetNetwork(partyNetConf);
```
GuutBoy: No reason to leave this comment in the code.

core/src/test/java/dk/alexandra/fresco/suite/dummy/bool/AbstractDummyBooleanTest.java
```diff
@@ -81,36 +81,35 @@ protected void runTest(
...
-       
+
 
       Drbg drbg = new HmacDrbg();
       TestThreadConfiguration<ResourcePoolImpl, ProtocolBuilderBinary> ttc =
           new TestThreadRunner.TestThreadConfiguration<>(sce,
               () -> new ResourcePoolImpl(playerId, noOfParties, drbg), () -> {
             Network network;
-            KryoNetNetwork kryoNetwork = new KryoNetNetwork(partyNetConf);
+            //network = new KryoNetNetwork(partyNetConf);
```
GuutBoy: No reason to leave this comment in the code.

suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/AbstractSpdzTest.java
```diff
@@ -99,13 +100,14 @@ protected void runTest(
               () -> createResourcePool(playerId, noOfParties, new Random(),
                   new SecureRandom(), preProStrat),
               () -> {
-                KryoNetNetwork kryoNetwork = new KryoNetNetwork(netConf.get(playerId));
+                //Network network = new KryoNetNetwork(netConf.get(playerId));
```
GuutBoy: No reason to leave this comment in the code

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
+import javax.net.ServerSocketFactory;
+import org.junit.Test;
+
+public class TestAsyncNetwork {
```
GuutBoy: General comments to the tests:

- I think some tests are missing here. It would be nice if the network is quite well tested. For example all tests here seem to be for 2 parties, and I do not see any tests of sending/receiving more than a single message. Maybe think about some of the ways the KryoNet implementation failed and add tests for that also. 
- It looks like this copies at lot from the KryoNet tests. It would be great to generalize the non-implementation specific tests into a set of tests which we can run on any new `Network` implementation.

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+
+  private List<Integer> getFreePorts(int no) {
```
GuutBoy: There is really not much guarantee that these ports are free. In particular if tests are run in parrallel, we definitely do not get free ports. Find a better strategy.

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+    List<Integer> res = new ArrayList<>();
+    for (int i = 0; i < no; i++) {
+      res.add(9000 + i);
+    }
+    return res;
+  }
+
+  @Test(timeout = 5000)
```
GuutBoy: Not sure if this is the right way to handle timeouts. On very slow machines this might just cause the test to fail. I think we are seeing that on Travis. Not sure what is right way to go though.

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+  public void testSendBytes() {
+    Map<Integer, Party> parties = new HashMap<>();
+    List<Integer> ports = getFreePorts(2);
+    parties.put(1, new Party(1, "localhost", ports.get(0), "78dbinb27xi1i"));
+    parties.put(2, new Party(2, "localhost", ports.get(1), "h287hs287g22n"));
```
GuutBoy: There is a lot of boiler-plate setup in these tests. Please refactor to make the tests more readable.

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+      res.add(9000 + i);
+    }
+    return res;
+  }
+
+  @Test(timeout = 5000)
+  public void testSendBytes() {
+    Map<Integer, Party> parties = new HashMap<>();
+    List<Integer> ports = getFreePorts(2);
+    parties.put(1, new Party(1, "localhost", ports.get(0), "78dbinb27xi1i"));
```
GuutBoy: We are not using the key part, so why specify it in the constructor?

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
```diff
@@ -0,0 +1,316 @@
...
+    network.close();
+    try {
+      t1.join(1000);
+    } catch (InterruptedException e) {
+      fail("Threads should finish without main getting interrupted");
+    }
+  }
+
+  @Test(timeout = 5000)
+  public void testSendBytesAllowMultipleMessages() {
```
GuutBoy: As far as I can see this test is essentially the same as the above.

core/src/test/java/dk/alexandra/fresco/framework/network/async/TestAsyncNetwork.java
GuutBoy: I.e., setting up parties and networks, could probably be factored out nicely for most of these tests.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,303 @@
...
+   * other parties.
+   *
+   * @param conf The network configuration
+   * @param timeout
+   */
+  public AsyncNetwork(NetworkConfiguration conf, int timeout) {
+    this.conf = conf;
+    this.clients = new HashMap<>();
+    this.queues = new HashMap<>();
+    int fixedSize = Math.max(1, getNoOfParties() - 1);
```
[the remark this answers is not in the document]
Rhaea: Assuming it was a "why": Since the dummy suites have a single party, and if the size is 0, exceptions will occur

GuutBoy:
I forgot one comment. Why are we using two connections for each pair of parties? The channel should be bidirectional, so one channel should be enough.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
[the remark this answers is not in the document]
GuutBoy: Sure, but then you do not need the receiverService at all, do you?

GuutBoy:
I made a few changes. Mainly, how connections are set up and changing to a single channel strategy. Also, added some more tests. As far as I can see this should be ready now.

core/src/main/java/dk/alexandra/fresco/framework/network/CloseableNetwork.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.io.Closeable;
+
+/**
+ * Interface for a closable network.
+ */
+public interface CloseableNetwork extends Network, Closeable {
+
+  /*
+   * Override the close method to NOT throw IOExceptions (this may be kind of a hack)
```
pffrandsen: ... hack?

Why circumvent it? If it is not closable, then do not inherit the interface

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,336 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(AsyncNetwork.class);
+
+  private ServerSocketChannel server;
+  private Map<Integer, SocketChannel> channelMap;
+  private Map<Integer, BlockingQueue<byte[]>> queues;
+  private NetworkConfiguration conf;
+  private ExecutorService receiverService;
+  private Map<Integer, ExecutorService> senderServices;
+  private Exception sendException = null;
+  private Exception receiveException = null;
```
pffrandsen: They are null by default?

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,336 @@
...
+    }
+  }
+
+  private void startReceivers() {
+    for (Entry<Integer, SocketChannel> entry : this.channelMap.entrySet()) {
+      final int id = entry.getKey();
+      SocketChannel channel = entry.getValue();
+      this.receiverService.submit(() -> {
+        try {
+          while (true) {
```
pffrandsen: while(true) and executorService is a poor match, either use a thread or resubmit. 
This would also change the structure of the exception propagation

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,336 @@
...
+          ByteBuffer b = ByteBuffer.allocate(1);
+          b.put((byte) conf.getMyId());
+          b.position(0);
+          channel.write(b);
+          connectionMade = true;
+          channelMap.put(i, channel);
+          logger.info("P{} connected to {}", conf.getMyId(), p);
+        } catch (IOException e) {
+          attempts++;
+          Thread.sleep(1 << attempts);
```
pffrandsen: Create a variable, timeout and call timeout=timeout*timeout. This is unbearable :)

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,336 @@
...
+        try {
+          ByteBuffer buf = ByteBuffer.allocate(Integer.BYTES + data.length);
+          buf.putInt(data.length);
+          buf.put(data);
+          buf.position(0);
+          while (buf.hasRemaining()) {
+            this.channelMap.get(partyId).write(buf);
+          }
+        } catch (Exception e) {
+          e.printStackTrace();
```
pffrandsen: No

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -0,0 +1,336 @@
...
+          this.sendException = e;
+        }
+        return null;
+      });
+      if (sendException != null) {
+        throw new RuntimeException("Previous send operation failed due to exception",
+            sendException);
```
pffrandsen: This exception handling should be centralized somewhere, maybe have an error handler in the class that can be signaled in case of an error.
Implicit handling through two variables with exceptions is unorthodox

pffrandsen: Submitted futures can (and should) be kept and checked for errors when sending

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
GuutBoy: Aw, I thought this was pretty elegant really.

core/src/main/java/dk/alexandra/fresco/framework/network/CloseableNetwork.java
GuutBoy: It is closeable. I just dont want to have to throw an IOException.

pffrandsen:
It is when bitlogik is close by. In this it is not - so you are forcing 
the reader to turn on the bitter. Which for some might be fine, however 
to me it seems like a cumbersome way of expressing the obvious

core/src/main/java/dk/alexandra/fresco/framework/network/CloseableNetwork.java
GuutBoy: ... I could just define my own interface and not inherit from Closeable, but I thought I would like to be able to use Classes implementing the interface in try-with-resources statements.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
GuutBoy: How would it work with resubmits? I cache the future and check if the latest future was cancelled in the receive method?

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
GuutBoy: "No" to printing the exception?

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
GuutBoy: I must admit I was not super happy with this solution. I guess I could save previous futures. That makes sense.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
pffrandsen: Well in the receive you only schedule the next receive if the current was successful and subsequently overwrite the "next" future.

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
pffrandsen: Yes, they need to be logged

pffrandsen: Oh darn. No, exceptions should be logged

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
pffrandsen: hmm 
github does not handle comments via mail properly
"It is when bitlogic is close by. In this it is not - so you are forcing
the reader to turn on the bitter. Which for some might be fine, however
to me it seems like a cumbersome way of expressing the obvious"

core/src/main/java/dk/alexandra/fresco/framework/network/CloseableNetwork.java
pffrandsen: Maybe use AutoCloseable, I mean the mere fact you are saying hack makes my nose cringe :)

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
GuutBoy: Will that work? Do I have a guarantee that the last receive is done when I submit the next?

pffrandsen: In this case (recieving) there is only one entity submitting, so yes

core/src/main/java/dk/alexandra/fresco/framework/network/CloseableNetwork.java
GuutBoy: AutoCloaseable is not better. It also throws checked exception on close :(.

GuutBoy: But ok, it's not a huge deal to just handle the IOException. Maybe, the more interesting question is why `Network` does not extend `Closeable` already?

PR #217, opened 2018-01-17
Fixed point arithmetic

MichaelStausholm:
Added a tool for doing fixed point arithmetic. Currently it does not support mixed (integer/fixedpoint) operations.


tools/fixedpointarithmetic/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+      <groupId>dk.alexandra.fresco</groupId>
+      <artifactId>core</artifactId>
+      <version>${fresco-version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <!-- Used for plaintext LP solving -->
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-math3</artifactId>
```
pffrandsen: Not needed?

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/DefaultFixedNumeric.java
```diff
@@ -0,0 +1,147 @@
...
+   * @param precision  Amount of digits after the dot.
+   */
+  public DefaultFixedNumeric(ProtocolBuilderNumeric builder, int precision) {
+    super();
+    this.precision = precision;
+    this.builder = builder;
+  }
+  
+  @Override
+  public DRes<SFixed> known(BigDecimal value) {
```
pffrandsen: Wondering if double suffices?

int is capped after 32 bits but double can represent a lot more with the 64 bits.

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/SFixed.java
```diff
@@ -0,0 +1,21 @@
...
+
+import dk.alexandra.fresco.framework.DRes;
+import dk.alexandra.fresco.framework.value.SInt;
+
+/**
+ * Wrapper for SInt used to represent fixed point values.
+ */
+public class SFixed {
+
+  private DRes<SInt> value;
```
pffrandsen: final

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/SFixed.java
```diff
@@ -0,0 +1,21 @@
+package dk.alexandra.fresco.fixedpoint;
+
+import dk.alexandra.fresco.framework.DRes;
+import dk.alexandra.fresco.framework.value.SInt;
+
+/**
+ * Wrapper for SInt used to represent fixed point values.
+ */
+public class SFixed {
```
pffrandsen: Similarly to the SInt this should be an interface and then there should be a single implementation called SFixedSIntWrapper

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/SFixed.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private DRes<SInt> value;
+  
+  public SFixed(DRes<SInt> value) {
+    this.value = value;
+  }
+  
+  public DRes<SInt> getSInt() {
```
pffrandsen: Rgiht now every user can get the underlying SInt value - that is not intented.

tools/fixedpointarithmetic/src/test/java/dk/alexandra/fresco/fixedpoint/basic/BasicArithmeticTests.java
```diff
@@ -0,0 +1,339 @@
...
+
+            return numeric.open(product);
+          };
+          BigDecimal output = runApplication(app);
+          Assert.assertEquals(value.divide(value2)
+              .setScale(5, RoundingMode.HALF_UP), output);
+        }
+      };
+    }
+  }
```
pffrandsen: Maybe rewrite a few of the tests to be a list of product/sums that are tested instead of just two single values

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/DefaultFixedNumeric.java
MichaelStausholm: It probably is, however then we need to manually handle scales and double <-> BigInteger conversion.

pffrandsen: Makes sense - let us keep it like this for now

tools/fixedpointarithmetic/pom.xml
MichaelStausholm: Removed.

pffrandsen:
You have merged e4cb1f2 - refactor-additive-mask - but it is not the same version as in the branch on github. Please make sure the other branch is merged properly and then correct the implementation here

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/FixedNumeric.java
```diff
@@ -0,0 +1,135 @@
+package dk.alexandra.fresco.fixedpoint;
+
+import dk.alexandra.fresco.framework.DRes;
+import dk.alexandra.fresco.framework.builder.ComputationDirectory;
+import java.math.BigDecimal;
+
+/**
+ * Basic interface for fixed point numeric applications.
+ */
+public interface FixedNumeric<SFixedT extends SFixed> extends ComputationDirectory {
```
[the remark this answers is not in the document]
MichaelStausholm: Right now we are not entirely sure how the fixedpoints will actually be used. The operations implemented currently are the ones we are fairly sure will be needed for some simple machine learning. If we need conversion, it is easy to implement (just a division or multiplication). The same goes for an "advanced" computation directory. If/as we get more operations, it might become relevant, but currently I believe it makes sense with a single fixed numeric computation directory.

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/SIntWrapperFixedNumeric.java
```diff
@@ -0,0 +1,148 @@
...
+  @Override
+  public DRes<SFixed> input(BigDecimal value, int inputParty) {
+    value = value.setScale(this.precision, RoundingMode.DOWN);
+    DRes<SInt> input = builder.numeric().input(value.unscaledValue(), inputParty);
+    return new SFixedSIntWrapper(input);
+  }
+
+  @Override
+  public DRes<BigDecimal> open(DRes<SFixed> secretShare) {
+    DRes<SInt> sint = ((SFixedSIntWrapper)secretShare.out()).getSInt();
```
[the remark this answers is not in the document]
MichaelStausholm: We want to hide the wrapped SInt from the app developer, but in order to make the generic types work, we have to use this workaround. It's not pretty, but it is consistent with the way we do this is Spdz.

tools/fixedpointarithmetic/src/main/java/dk/alexandra/fresco/fixedpoint/FixedNumeric.java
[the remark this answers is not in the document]
Rhaea: My point was also that since it is easy to implement conversions, we might as well do it, but I guess we're going for the principle of: are we using it? If not, throw it away, so implementing something that might not be used is probably a bad idea now that I think twice.

PR #223, opened 2018-01-18
Codestyle

MichaelStausholm:
I got (really) bored on the train trip from Copenhagen, so I made some codestyle fixes in core. 

This should not alter functionality and is part of issue #114 .


core/src/main/java/dk/alexandra/fresco/framework/sce/evaluator/BatchedProtocolEvaluator.java
```diff
@@ -73,12 +73,12 @@ public void eval(ProtocolProducer protocolProducer, ResourcePoolT resourcePool,
       } else {
         zeroBatches = 0;
       }
-/*
+      /*
```
Rhaea: We should look into this issue at some point. Maybe make an issue on it. It should be a bug if empty batches appear often, but apparently this has been removed for a reason. Should be dealt with somehow (not in this PR)

PR #224, opened 2018-01-18
Refactor additive mask

MichaelStausholm:
The AdvancedNumeric.additiveMask method automatically added a magic security parameter to the AdditiveMask constructor. This responsibility is now moved to the code calling additiveMask(int noOfBits). 

Hence additiveMask can now be used to get a noOfBits size random element.

Should we rename additiveMask to random(int bitSize) ?


pffrandsen:
Right now the magic number is used three places,
- LessThan
- ZeroTestReducer
- RightShift
Why is it used these places?

I am assuming the 60 is the sufficient for the comparison to run with a reasonable likelihood, however why then the RightShift?

Another issue is the fact that it is now hardcoded into the computation builder and therefore hard to change, please make it a construction parameter to the relevant computations (i.e. less than et al)

MichaelStausholm:
I'm not really sure where Jonas got 60, however a 0 parameter value would leak information in all 3 cases from what I understand from Jonas.

I added it as a construction parameter, however the magicnumer nows resides in both DefaultAdvancedBuilder and DefaultComparison. Is there a more natural place to store it?

PR #233, opened 2018-02-01
221 fix correlation error bug

### LINKED ISSUE #221: Fix correlation error-bug
When running MASCOT triple generation on Partisia's machines (or specifically @pffrandsen machine), with at least 5 instances in parallel, an error occurs. A MaliciousException is thrown from the OT extension. Specifically from RotSender.java stating that "Correlation check failed for the sender in the random OT extension".
This has not been reproducible on OS X machines when using KryoNet. Note that Peter's machine is slow, runs Windows and uses the Memory Network and not KryoNet.

pffrandsen:
Closes #221


GuutBoy:
Nice catch 👍

pffrandsen:
Honor goes to @eorloff

PR #234, opened 2018-02-02
Faster repeated right shift protocol

jonas-lj:
Instead of calling the right shift protocol multiple times to do multiple right shift, we can do it in one go at the cost of a multiplication per extra shift.


GuutBoy:
@jonas-lj something may have gone wrong in the conflict resolution here. This is no longer compiling.

jonas-lj:
I forgot to remove the fixed-pt module from the tools pom, but it should be fixed now.

GuutBoy:
I think it is a different issue for this PR.

PR #238, opened 2018-02-15
Addressing issue 231 (cleaning up drbg)

n1v0lg:
Added default constructor to AesCtrDrbg and moved drbg away from ResourcePool to resource pools that actually require this functionality (Mascot and OTExtension).


core/src/test/java/dk/alexandra/fresco/framework/util/TestAesCtrDrbg.java
```diff
@@ -5,10 +5,25 @@
 
 import java.util.Arrays;
 import java.util.Random;
+import org.hamcrest.core.IsEqual;
+import org.hamcrest.core.IsNot;
+import org.junit.Assert;
```
GuutBoy: I prefer static imports for these assertions, but really it is a matter of taste.

tools/mascot/src/main/java/dk/alexandra/fresco/tools/mascot/MascotResourcePool.java
```diff
@@ -92,4 +93,15 @@
...
+
+  /**
+   * The DRBG is useful for protocols which needs a form of shared randomness where the random bytes
+   * are not easily guessed by an adversary. This generator will provide exactly that. For explicit
+   * security guarantees, we refer to implementations of
+   * {@link dk.alexandra.fresco.framework.util.Drbg}.
+   *
+   * @return An instance of a DRBG.
+   */
+  Drbg getRandomGenerator();
```
GuutBoy: Is the documentation still accurate for the use in MASCOT?

PR #239, opened 2018-02-15
Spdz maccheck as computation

pffrandsen:
After discussing the protocol producers with n1v0lg and having seen his very elegant example in Marlin, I rewrote the mac check to be based on computations. 
This also obliterates the need for protocol producing collections


suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -36,95 +38,80 @@ public SpdzMacCheckProtocol(
...
-        rs[i] = temporaryR;
-      }
-      BigInteger a = BigInteger.ZERO;
-      int index = 0;
-      for (BigInteger aa : openedValues) {
-        a = a.add(aa.multiply(rs[index++])).mod(modulus);
-      }
+          // Add all s's to get the common random value:
+          BigInteger sum = BigInteger.ZERO;
+          for (BigInteger otherS : openCommit.values()) {
```
GuutBoy: How old fashioned ... Why not `openCommit.stream.reduce(ZERO, BigInteger::add);`

PR #243, opened 2018-02-23
Add logo to readthedocs and various minor fixes.

### LINKED ISSUE #242: Add new logo to readthedocs
Add the new logo to our readthedocs site.

GuutBoy:
Adds the new logo to the front page. Also, fixes multiple minor typos and rephrases the docs a bit.

This fixes #242

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -3,35 +3,42 @@
 Contributing
 ============
 
-We will be happy accept contributions to the FRESCO framework.  
+We will be happy to accept contributions to the FRESCO framework. Contributions could include:
 
-* **Feedback.** Whether you have a *question* on FRESCO, a *suggestion* for new additions or
-  noticed a *bug* in the framework, any feedback is a valuable contribution to our work on FRESCO.
+* **Feedback** - Whether you have a *question* on FRESCO, a *suggestion* for new additions or
+  noticed a *bug* in the framework, any feedback is a valuable contribution.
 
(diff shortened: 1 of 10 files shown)


doc/source/contributing.rst
```diff
@@ -3,35 +3,42 @@
 Contributing
```
Rhaea: Is this file needed now that you included contributing.md? Seems like it's duplicate text which I feel the same about as with code. Not a huge issue, but they might not be identical in the future.

doc/source/index.rst
```diff
@@ -3,14 +3,21 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
   
-
-FRESCO
-======
+.. image:: https://alexandra.dk/logo/fresco-logo-artwork.svg
```
Rhaea: Wheee! :D

doc/source/contributing.rst
GuutBoy: I totally get your point, and I tend to agree. However, I am not really sure where I think this information mainly belongs, so for now I would like to keep it this way. I mainly put the `CONTRIBUTING.md` files because it fits in to the whole GitHub experience :). 

If we find that we want to change these files often then we should just pick a single location though.

GuutBoy:
Good point about adding a description of how to get version 1.0.0 from the central repo. I will add that.

GuutBoy:
@Rhaea how do you like that change to the install?

PR #249, opened 2018-03-15
Minimize drbg initialization in ot extension, small optimizations and streams

### LINKED ISSUE #248: Minimize DRBG initialization in MASCOT/OTE
As mentioned in #218 the MASCOT and OT extension implementations is initializing too many DRBG's where this is not strictly needed. To optimize we should minimize the amount of initializations.

GuutBoy:
This pull request:
1. Reduces the number of DRBG initializations in the `BristolRotBatch` class in order to optimize OT extension performance. This is done by introducing a utility class to adjust the length a random string to match a desired length. This works by, if necessary, stretching the string using a hash function but otherwise using the provided string. An other utility uses this to implement form of OTP encryption. (fixes #248 )
2. Moves many parts of the OT-extension from loops to streams.
3. Improves the `Transpose` utility class by copying byte arrays bytewise instead of bitwise
4. Improves the `MultiplyWithoutReduction` method in the `RotSharedImpl` class by pre-computing all possible rotations of the `bvec` input.


GuutBoy:
@pffrandsen I totally agree, this branch got a little out of hand. I will do better in the furture. I also agree the performance gain is a little disappointing. One way to improve might be to use more parallelization. The addition of streams should make this quite easy to add.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/NaorPinkasOt.java
```diff
@@ -27,53 +25,38 @@
...
-    this.p = params.getP();
-    this.g = params.getG();
+    this.dhModulus = params.getP();
+    this.dhGenerator = params.getG();
     this.randNum = new DrngImpl(randBit);
   }
 
   @Override
   public void send(StrictBitVector messageZero, StrictBitVector messageOne) {
-    // To avoid leakage both messages sent will have the length of the longest
```
jot2re: Why are you removing the comment?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/BristolRotBatch.java
```diff
@@ -42,72 +37,51 @@ public BristolRotBatch(RotFactory randomOtExtension, int comSecParam, int statSe
...
-    StrictBitVector extraChoices = new StrictBitVector(extraByteChoices);
-    List<StrictBitVector> messages = receiver.extend(extraChoices);
-    for (int i = 0; i < choiceBits.getSize(); i++) {
-      StrictBitVector newMessage = computeRandomMessage(messages.get(i),
-          sizeOfEachMessage);
-      res.add(newMessage);
-    }
-    return res;
+    int amountToPreprocess = computeExtensionSize(choiceBits.getSize(), comSecParam, statSecParam);
+    byte[] extraByteChoices = Arrays.copyOf(choiceBits.toByteArray(), amountToPreprocess / 8);
```
jot2re: Shouldn't it be  Byte.SIZE instead of 8, just to be consistent in the usage of constants :)

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/LengthAdjustment.java
```diff
@@ -0,0 +1,79 @@
...
+    return digest;
+  }
+
+  /**
+   * Generates a byte array representation of an integer.
+   *
+   * @param i an integer
+   * @return a corresponding byte array
+   */
+  private static byte[] intToBytes(int i) {
```
jot2re: I guess that for (rather pedantic) consistency we should do a loop from 0 to Integer.BYTES and shift based on a multiple of Byte.BIT,  rather than having hardcoded 4 bytes for an int and 8 bits per byte. But I think that might be a bit too much to do, because we already have the necessary guarantees from Java that what is below will work.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/LengthAdjustmentTest.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  private void testShorterCandidateLength(final int candidateLength, final int adjustedLength) {
+    byte[] candidate = new byte[candidateLength];
+    new Random().nextBytes(candidate);
+    byte[] adjusted1 = LengthAdjustment.adjust(candidate, adjustedLength);
+    assertEquals(adjustedLength, adjusted1.length);
+    byte[] adjusted2 = LengthAdjustment.adjust(candidate, adjustedLength);
+    assertArrayEquals(adjusted1, adjusted2);
```
jot2re: Perhaps add sanity test that random output looks random, e.g. not the zero-vector.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/PseudoOtpTest.java
```diff
@@ -0,0 +1,134 @@
...
+    byte[] candidate = new byte[candidateLength];
+    rand.nextBytes(candidate);
+    byte[] cipherText = PseudoOtp.encrypt(message, candidate, cipherLength);
+    assertEquals(cipherLength, cipherText.length);
+    if (cipherText.length > 0) {
+     assertFalse(Arrays.equals(Arrays.copyOf(message, cipherLength), cipherText));
+     assertFalse(Arrays.equals(Arrays.copyOf(candidate, cipherLength), cipherText));
+    }
+    byte[] decryptedMessage = PseudoOtp.decrypt(cipherText, candidate, cipherLength);
+    assertArrayEquals(Arrays.copyOf(message, cipherLength), decryptedMessage);
```
jot2re: Perhaps we should also check that no zero-vectors have sneaked in here?

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/base/TestFunctionalNaorPinkas.java
```diff
@@ -48,23 +48,18 @@ public void shutdown() {
...
-    Network network = new CheatingNetwork(
-        RuntimeForTests.defaultNetworkConfiguration(1, Arrays.asList(1, 2)));
+  private List<Pair<StrictBitVector, StrictBitVector>> otSend(int iterations) throws Exception {
+    Network network =
+        new CheatingNetwork(RuntimeForTests.defaultNetworkConfiguration(1, Arrays.asList(1, 2)));
     try {
       Drbg rand = new AesCtrDrbg(HelperForTests.seedOne);
-      DhParameters params = new DhParameters();
-      Ot otSender = new NaorPinkasOt(2, rand, network, params
-          .computeSecureDhParams(1, 2, rand, network));
```
jot2re: Do we want to remove the option of computing computeSecureDhParams from DhParameters? Because if not, then the testing using this method should not have been removed from all the tests in this class. On the other hand if we do want to remove the computeSecureDhParams  option then the code for this method should be removed from DhParameters.

GuutBoy:
@jot2re I wrote some point by point comments below:

### Removing comments in NaorPinkasOt
I felt that the code was already nicely documented by having code factored into well named local methods, which were all well documented. So the comments did not seem to be needed. My personal taste is that in this case the code can actually be made harder to read by being cluttered by too many inline comments. I guess it is a matter of taste. Is there anything in particular where you feel the comments should be added back?

### Stream overhead
I think you are right that using streams can have bit of overhead, but as far as I can see it is not significant here. I really like the new stream API, and to me it makes the code very easy to understand, which is part of the reason why I rewrote it in that style. An other reason is that it makes it very easy to parallelize the computation (basically just add `.parallel()` to the pipeline). I did not do that in this PR since I was not sure how that would effect how we use MASCOT in the larger picture though (and as Peter pointed out I was already doing to much in this PR). 

### Transpose    
Yes, I mainly rewrote this to streams to make it more clear to myself. Once I reduced the number of DRBG initializations, it turned out transpose was the new bottleneck. To fix that I first "cleaned up" the code a bit to get a better idea of what was going on. As far as I can see transposing is no longer a bottleneck btw., so I think we should keep Eklundh. The new bottleneck appears to be `RotSharedImpl.multiplyWithoutReduction(...)`, which is why I did some minor optimizations on that method. However, it remains the main bottleneck in my local tests, so any suggestions to improve would be welcome 😄 (in c++ I think you can take advantage of the CMUL instructions, but not sure how to do that in Java). 

### Coverage
I agree. I guess I just focused on the patch having 100% coverage. I think, however, that I may have decreased DhParameters coverage by using static parameters in the tests (which I would like to keep doing because it really speeds up the tests though.).  

### Linelength
100 is the correct length https://google.github.io/styleguide/javaguide.html#s4.4-column-limit

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/RotSharedImpl.java
```diff
@@ -35,86 +34,73 @@ public RotSharedImpl(CoinTossing ct, MessageDigest digest, int comSecParam) {
...
-      res.xor(temp);
-    }
+  protected static StrictBitVector computeInnerProduct(List<StrictBitVector> alist,
+      List<StrictBitVector> blist) {
+    // All elements of each list MUST have equal size so we find the size using the first element.
+    StrictBitVector res = new StrictBitVector(alist.get(0).getSize() + blist.get(0).getSize());
+    List<StrictBitVector> products = IntStream.range(0, alist.size())
+        .mapToObj(i -> multiplyWithoutReduction(alist.get(i), blist.get(i)))
+        .collect(Collectors.toList());
+    res = products.stream().reduce(res, (a, b) -> {
```
n1v0lg: I'd remove the extra collect step, i.e., call reduce on the stream resulting from mapToObj.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/base/TestFunctionalNaorPinkas.java
GuutBoy: Good point, I will go with just removing the computeSecureDhParams method, as we are never really using it anyway.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/PseudoOtpTest.java
GuutBoy: I think that is already covered by testing for equality of the original message.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/RotSharedImpl.java
GuutBoy: Good point.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/LengthAdjustmentTest.java
GuutBoy: Done

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/LengthAdjustment.java
GuutBoy: I considered it! But I felt it would be simpler this way.

GuutBoy:
@pffrandsen The test in  [TestDhParameters](https://github.com/aicis/fresco/blob/555e8339bde2b7903d6577b16c3330ecd180bc80/tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/base/TestDhParameters.java) generates DhParameters in order to test against the static parameters. Is that all you are asking for?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/otextension/LengthAdjustment.java
jot2re: I must admit that I agree :)

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/PseudoOtpTest.java
jot2re: I agree, handled with the extra check in the lengthAdjustmentTest

jot2re:
@GuutBoy Some feedback below:

**Removing comments in NaorPinkasOt**

It is fine. No need to put things back. I was just curious about the motivation :) 

**Stream overhead**

If it doesn't have a significant impact then everything is fine. I have not used the streams in java before, but if it is that easy to parallelise using them, then it sounds like a great tool I/we should use much more in the future.

**Transpose**

Ok. I don't have any optimisation suggestions for RotSharedImpl.multiplyWithoutReduction(...) unfortunately.

**Coverage**

Looks good now :)

**Linelength**
Sorry. my fault.

PR #250, opened 2018-03-20
Optimize default expipe

GuutBoy:
Reduces the number of rounds required for the default exppipe computation from linear to logarithmic in the length of the pipe.


core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/DefaultPreprocessedValues.java
```diff
@@ -22,29 +23,29 @@ public DefaultPreprocessedValues(ProtocolBuilderNumeric builder) {
...
+        List<DRes<SInt>> newValues = state.values.stream().skip(1).limit(limit)
+            .map(v -> par.numeric().mult(last, v)).collect(Collectors.toList());
+        state.values.addAll(newValues);
+        return () -> new IterationState(state.values);
+      });
+
+    }).seq((seq, state) -> () -> state.values);
   }
 
   private static final class IterationState {
```
pffrandsen: This is not needed anymore, remove it

GuutBoy: Which part?

pffrandsen: ...

Trickster.

The entire class IterationState and can be replaced with the list

GuutBoy: Oh, you mean the iterationstate class? Sure I will get rid of that.

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/DefaultPreprocessedValues.java
```diff
@@ -22,29 +23,19 @@ public DefaultPreprocessedValues(ProtocolBuilderNumeric builder) {
...
-
-    private IterationState(int round, List<DRes<SInt>> value) {
-      this.round = round;
-      this.value = value;
-    }
+      ArrayList<DRes<SInt>> values = new ArrayList<>(pipeLength + 2);
+      values.add(inverse);
+      values.add(r);
+      return () -> values;
+    }).whileLoop((values) -> values.size() <= pipeLength + 1, (seq, values) -> {
```
pffrandsen: And now the small print
Could you write < rather than <= and then drop the +1?

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/DefaultPreprocessedValues.java
```diff
@@ -22,29 +23,19 @@ public DefaultPreprocessedValues(ProtocolBuilderNumeric builder) {
...
+      return seq.par(par -> {
+        DRes<SInt> last = values.get(values.size() - 1);
+        int limit = pipeLength + 2 - values.size();
+        List<DRes<SInt>> newValues = values.stream().skip(1).limit(limit)
+            .map(v -> par.numeric().mult(last, v)).collect(Collectors.toList());
+        values.addAll(newValues);
+        return () -> values;
+      });
+    }).seq((seq, values) -> () -> values);
```
pffrandsen: redundant

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/DefaultPreprocessedValues.java
GuutBoy: No I could not. Because that would not be equivalent to what is there now. I could go with `... values.size() < pipeLength + 2 ... ` and save a character.

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/DefaultPreprocessedValues.java
```diff
@@ -20,22 +21,45 @@ public DefaultPreprocessedValues(ProtocolBuilderNumeric builder) {
...
         return () -> values;
       });
-    });
+    } else {
+      return builder.seq(b -> {
+        DRes<SInt> r = b.numeric().randomElement();
+        DRes<SInt> inverse = b.advancedNumeric().invert(r);
+        List<DRes<SInt>> values = new ArrayList<>(pipeLength + 2);
+        values.add(inverse);
+        values.add(r);
```
pffrandsen: It seems as if you are producing values twice if pipelength>=2

GuutBoy: Good catch ... that was the whole point of the `firstStep` part 👍

PR #257, opened 2018-03-23
Full test coverage of ByteArrayHelper by making it not instantiable

GuutBoy:
Got annoyed that this class was not fully covered.


GuutBoy:
I realize this is not at 100% test coverage, but this should be fixed in the Ring Proto PR. Sorry, about that.

PR #264, opened 2018-04-05
fix AesDemo mainClass in pom.xml

enricobacis:
The wrong capitalization in `aes/pom.xml` prevents the generated jar from running on case-sensitive file-systems (i.e., Linux).


GuutBoy:
Thank you for spotting this 👍

Issue #265, opened 2018-04-05
Add Mascot to demos


n1v0lg:
The demos should include an example of how to use the Mascot-based data supplier together with SPDZ.

enricobacis:
I have something working here: https://github.com/enricobacis/fresco/tree/mascot-cli

The sum demo already works with MASCOT

jonas-lj:
This was fixed with PR #351

PR #267, opened 2018-04-08
Mascot cli

enricobacis:
Add MASCOT preprocessing strategy to the Demo Cli.
The sum example can now be run with MASCOT.

This should address #265 .


enricobacis:
I used the wrong issue number, I wanted to say that this should address #265

Issue #266, opened 2018-04-07
Mascot data supplier input mask generation error 


n1v0lg:
SpdzMascotDataSupplier does not correctly generate input masks when input is provided by more than one party. This is due to an error in the batched generation mechanism in the supplier. First observed by @enricobacis.

GuutBoy:
Closed in #268

PR #268, opened 2018-04-09
266 mascot input mask fix

n1v0lg:
Addresses issue #266. The mascot data supplier generates pre-processing material in batches and caches the results; this works correctly for party-independent material such as multiplication triples but does not work (w/o modification) for party-dependent input masks. This fix updates the mechanism for pre-generating and caching input masks per party (as opposed to in a single collection).


enricobacis:
Hi @n1v0lg, I tried these changes in #267 and I confirm that now also `distance` and `aggregation` demos are correctly working! I updated #267 with Makefile targets to run MASCOT preprocessing also for these other two demos.

PR #273, opened 2018-04-25
Introduced the SInt values for numerator and denominator in LP

pffrandsen:
Allows to sort and rank the results post computation


core/src/main/java/dk/alexandra/fresco/lib/statistics/DeaSolver.java
```diff
@@ -22,9 +22,6 @@
 import java.util.stream.IntStream;
 
 /**
- * NativeProtocol for solving DEA problems.
- *
- * <p>
  * Given a dataset (two matrices of inputs and outputs) and a number of query vectors, the protocol
```
GuutBoy: Shouldn't there be a `<p>` tag here?

pffrandsen: I removed the header stating the wrong intention.

PR #285, opened 2018-05-03
Replace kryo network and fix newly uncovered AsyncNetwork problems

### LINKED ISSUE #281: Replace KryoNet Network with AsyncNetwork
In order to be able to remove KryoNet dependency we need to replace all usage of the KryoNet Network with the new `AsyncNetwork` implementation.

GuutBoy:
This pull request removes and replaces the KryoNet based network implementation. Doing so uncovered some unfortunate behavior in the AsyncNetwork implementation which this PR also fixes and the AsyncNetwork is cleaned up a bit.  

This closes #281


suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/TestMaliciousBehaviour.java
```diff
@@ -52,73 +51,53 @@ public void reset() {
...
-  @Test
-  public void testMaliciousInput() {
+  @Test(expected=MaliciousException.class)
+  public void testMaliciousInput() throws Throwable {
     try {
       runTest(new BasicArithmeticTests.TestInput<>(), EvaluationStrategy.SEQUENTIAL_BATCHED, 2,
           Corrupt.INPUT);
-      Assert.fail("Should not go well");
     } catch (RuntimeException e) {
-      if (e.getCause().getCause() == null || !(e.getCause()
```
pffrandsen: though prettier, does not have a the same sematic

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/CheatingNetwork.java
```diff
@@ -1,27 +1,31 @@
...
+  private final CloseableNetwork network;
+
   private int cheatByteNo;
   private int messageNo;
   private int counter = 0;
   private boolean cheat = false;
 
   public CheatingNetwork(NetworkConfiguration conf) {
-    super(conf);
+    this.network = new AsyncNetwork(conf);
```
pffrandsen: Since it now has become a decorator, then call the network decorated or delegate

core/src/main/java/dk/alexandra/fresco/framework/network/async/AsyncNetwork.java
```diff
@@ -26,7 +28,6 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
```
pffrandsen: This class is now the only implementation of network, move it to the network package instead of ...async

suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/TestMaliciousBehaviour.java
GuutBoy: True, but the test will fail/pass on the same cases. And it is much more clear what we are testing for now.

tools/ot/src/test/java/dk/alexandra/fresco/tools/ot/otextension/CheatingNetwork.java
GuutBoy: Good point ... in fact, think to should not even be dependent on AsyncNetwork. We could use any closeable network here.

suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/TestMaliciousBehaviour.java
pffrandsen: If the latter statement was true, then the emantic is identical. Which it is not. Call me for explanation.

As I said, it is prettier so this is just a comment. Do not change anything

GuutBoy:
@pffrandsen if you have ideas on how to make it better/more failsafe, I would be happy to know and update it again. 

IMO the main gain of this implementation is to have a simple implementation that we understand well, making it easier to fix bugs or adjust the implementation we want/need. So as long at it is otherwise not considerably worse than the KryoNet based implementation (which it does not seem to be), I think it is preferable.

pffrandsen:
@GuutBoy It is not because I have ready ideas. My gut feeling is the process has not been adequately controlled in order to produce failsafe code. It does seem like trial and error. Hence my comment. So this remark really is stating: Hmm

PR #287, opened 2018-05-08
Lpsolver update with max iterations

pffrandsen:


core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -85,32 +88,47 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
...
         if (pivotRule == PivotRule.BLAND) {
           return blandPhaseOneProtocol(inner, state);
         } else {
           return phaseOneProtocol(inner, state, zero);
         }
       }).seq((inner, phaseOneOutput) -> {
-        if (!phaseOneOutput.terminated()) {
-          phaseTwoProtocol(inner, phaseOneOutput);
+        BigInteger isFinished = phaseOneOutput.getSecond().out();
+        if (isFinished.equals(BigInteger.ONE)) {
```
GuutBoy: This sequence of if statements seems overly complex. Maybe it would help to use, say, BigInteger.TWO to signal the abort state?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -85,32 +88,47 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
...
+          if (phaseOneOutput.getFirst() != null) {
+            LpState lpState = new LpState(state.tableau, state.updateMatrix, state.basis,
```
GuutBoy: Any reason not just to return the state? If we need to copy it like this maybe a copy constructor would be more clear?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -85,32 +88,47 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
...
+                state.pivot, state.iteration);
+            return () -> lpState;
+          } else {
+            return LpState::new;
```
GuutBoy: Why not have a constant LpState.ABORT_STATE to return here?

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -170,7 +187,8 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
    *
    * @return a protocol producer for the first half of a simplex iteration
    */
-  private DRes<LpState> phaseOneProtocol(ProtocolBuilderNumeric builder, LpState state,
+  private DRes<Pair<List<DRes<SInt>>, DRes<BigInteger>>> phaseOneProtocol(
```
GuutBoy: The new return type is not reflected in the javadoc ... of course the old one wasn't either :(

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -198,7 +215,8 @@ private boolean checkDimensions(LPTableau tableau, Matrix<DRes<SInt>> updateMatr
    *
    * @return a protocol producer for the first half of a simplex iteration
    */
-  private DRes<LpState> blandPhaseOneProtocol(ProtocolBuilderNumeric builder, LpState state) {
+  private DRes<Pair<List<DRes<SInt>>, DRes<BigInteger>>> blandPhaseOneProtocol(
```
GuutBoy: As above

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -281,37 +297,50 @@ public LPOutput(LPTableau tableau, Matrix<DRes<SInt>> updateMatrix, List<DRes<SI
...
+    public LpState(LPTableau tableau, Matrix<DRes<SInt>> updateMatrix,
+        List<DRes<SInt>> basis, DRes<SInt> pivot, int iteration) {
+      // Phase one completed
+      this(basis, updateMatrix, pivot, iteration, tableau, null, true, pivot);
     }
 
-    @Override
-    public LpState out() {
-      return this;
+    public LpState() {
```
GuutBoy: Any reason not to have a `public static final LpState ABORT_STATE = new LpState(null, null, null, null, -1);` or similar? It seems it would be more clear to me.

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
```diff
@@ -281,37 +297,50 @@ public LPOutput(LPTableau tableau, Matrix<DRes<SInt>> updateMatrix, List<DRes<SI
...
+    private final DRes<SInt> pivot;
+    private final List<BigInteger> enumeratedVariables;
+    private final List<DRes<SInt>> basis;
+    private final DRes<SInt> prevPivot;
 
-    public LpState(BigInteger terminationOut, LPTableau tableau, Matrix<DRes<SInt>> updateMatrix,
-        List<DRes<SInt>> enteringIndex, DRes<SInt> pivot, List<BigInteger> enumeratedVariables,
-        List<DRes<SInt>> basis, DRes<SInt> prevPivot) {
-      this.terminationOut = () -> terminationOut;
+    public LpState(List<DRes<SInt>> basis, Matrix<DRes<SInt>> updateMatrix,
```
GuutBoy: The constructors of the LpState are pretty complex. We should at least sort the arguments in the same order in each constructor.

core/src/main/java/dk/alexandra/fresco/lib/statistics/DeaSolver.java
```diff
@@ -133,9 +137,15 @@ private boolean consistencyCheck() {
         List<DRes<SInt>> initialBasis = prefix.getBasis();
 
         result.add(par.seq(subSeq -> subSeq.seq((solverSec) -> {
-          LPSolver lpSolver = new LPSolver(pivotRule, tableau, update, pivot, initialBasis);
+          LPSolver lpSolver = new LPSolver(
+              pivotRule, tableau, update, pivot, initialBasis, maxNumberOfIterations);
           return lpSolver.buildComputation(solverSec);
         }).seq((optSec, lpOutput) -> {
+          if (lpOutput.pivot == null) {
```
GuutBoy: Is this how we check for an aborted LpSolver? Maybe make that state more clear?

core/src/test/java/dk/alexandra/fresco/lib/lp/LpBuildingBlockTests.java
```diff
@@ -505,7 +506,7 @@ public void test() {
    * @param value the value
    */
   private static void assertDebugInfoContains(String output, String key, String value) {
-    assertTrue(output.contains(key + ": \n" + value));
+    Assert.assertThat(output, Matchers.containsString(key + ": \n" + value));
```
GuutBoy: For consistency, please use a static import for `assertThat` here for consistency.

core/src/main/java/dk/alexandra/fresco/lib/lp/LPSolver.java
pffrandsen: Mutable state in a while loop is not ideal, I will look at copying

PR #292, opened 2018-07-05
Replaces the padding drbg

### LINKED ISSUE #291: Replace PaddingAesCtrDrbg
The  `PaddingAesCtrDrbg` is used to get an `AesCtrDrbg` from a seed of length less than 32 bytes, by simply padding to length with zeroes. This is not a recommended way to derive a seed. 

We should replace this with a method that does something more sensible to derive the the seed (i.e., just hash it). Might also be nice to be able to get a DRBG from a seed longer than 32 bytes while we are at it.

GuutBoy:
This replaces the PaddingAesCtrDrbg with a factory class for the AesCtrDrbg class. 

The PaddingAesCtrDrbg class was used when a seed was not exactly the 32 byte seeds required by the AesCtrDrbg. It was simply a wrapper around the AesCtrDrbg but with a constructor that allowed a shorter seed by padding a short seed to length with zeros and calling the AesCtrDrbg. This however, is not really a great way to derive a seed.
 
The new factory provides factory methods for the existing AesCtrDrbg plus a method to construct the AesCtrDrbg from a derived seed. This will take any length byte array and hash it (using SHA256) to a 32 byte seed which is then used in the usual constructor.

This closes #291


core/src/main/java/dk/alexandra/fresco/framework/util/AesCtrDrbgFactory.java
```diff
@@ -0,0 +1,74 @@
...
+   * Note: This method should be used with caution as, depending on the context, a seed derived
+   * from a low entropy sequence of bytes can be insecure to use.
+   * </p>
+   *
+   * @param bytes a sequence of bytes from which to derive the seed
+   * @return a new DRBG
+   * @throws NoSuchAlgorithmException if the hash algorithm {@value #HASH_ALGORITHM} is not
+   *      available on the system
+   */
+  public static Drbg fromDerivedSeed(byte... bytes) throws NoSuchAlgorithmException {
```
pffrandsen: It is annoying that this method is typed with an exception that cannot be controlled by the caller. 

I suggest you wrap it with ExceptionConverter.safe so callers do not have to handle a broken jvm

suite/spdz/src/test/java/dk/alexandra/fresco/suite/spdz/storage/TestParallelMascots.java
```diff
@@ -112,7 +112,9 @@ public void testConstructor() throws Exception {
   private Drbg getDrbg() {
     byte[] drbgSeed = new byte[mascotSecurityParameters.getPrgSeedLength() / 8];
     new SecureRandom().nextBytes(drbgSeed);
-    return new PaddingAesCtrDrbg(drbgSeed);
+    Drbg drbg = ExceptionConverter.safe(() ->
+        AesCtrDrbgFactory.fromDerivedSeed(drbgSeed), "Could not generate drbg.");
```
pffrandsen: Especially since all the callers now have to handle it...

core/src/main/java/dk/alexandra/fresco/framework/util/AesCtrDrbgFactory.java
GuutBoy: Your right, I will change this.

core/src/main/java/dk/alexandra/fresco/framework/util/AesCtrDrbgFactory.java
```diff
@@ -0,0 +1,75 @@
...
+   * Note: This method should be used with caution as, depending on the context, a seed derived
+   * from a low entropy sequence of bytes can be insecure to use.
+   * </p>
+   *
+   * @param bytes a sequence of bytes from which to derive the seed
+   * @return a new DRBG
+   * @throws NoSuchAlgorithmException if the hash algorithm {@value #HASH_ALGORITHM} is not
+   *      available on the system
+   */
+  public static Drbg fromDerivedSeed(byte... bytes) {
```
n1v0lg: bytes could be final.

core/src/main/java/dk/alexandra/fresco/framework/util/AesCtrDrbgFactory.java
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * <p>
+   * Note: the seeds should be an array of exactly {@link AesCtrDrbg#SEED_LENGTH} uniformly random
+   * bytes. If the length constraint is not met an {@link IllegalArgumentException} will be thrown.
+   * </p>
+   *
+   * @param seed {@link AesCtrDrbg#SEED_LENGTH} uniformly random bytes
+   * @return a new DRBG
+   */
+  public static Drbg fromRandomSeed(byte[] seed) {
```
n1v0lg: seed could be final

core/src/main/java/dk/alexandra/fresco/framework/util/AesCtrDrbgFactory.java
GuutBoy: I suppose so, I never really got when method parameters should be declared final. But it kind of makes sense here.

PR #305, opened 2018-08-22
Exponentiation bug

### LINKED ISSUE #303: Exponentiation with open exponent fails when exponent is a 2-power
A bug in `ExponentiationOpenExponent.java` means that a `NullPointerException` is thrown when computing exponentiation with an exponent that is a power of 2. 

The problem arises as the `IterationState` object is not initialized properly.

GuutBoy:
Fixes a bug pointed out on gitter where the exponentiation with and open exponent would fail with a `NullPointerException` when the exponent was a power of two. 

This closes #303

### DIFF
--- a/core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
+++ b/core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
@@ -8,7 +8,7 @@ import dk.alexandra.fresco.framework.value.SInt;
 import java.math.BigInteger;
 
 /**
- * Computes the exponentiation when the exponent is public. 
+ * Computes the exponentiation when the exponent is public.
  */
 public class ExponentiationOpenExponent implements
     Computation<SInt, ProtocolBuilderNumeric> {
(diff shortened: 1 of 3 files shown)


core/src/main/java/dk/alexandra/fresco/lib/math/integer/exp/ExponentiationOpenExponent.java
```diff
@@ -29,7 +36,8 @@ public ExponentiationOpenExponent(DRes<SInt> x, BigInteger e) {
   public DRes<SInt> buildComputation(ProtocolBuilderNumeric builder) {
     return builder.seq((seq) -> {
       DRes<SInt> accEven = base;
-      return new IterationState(exponent, accEven, null);
+      DRes<SInt> accOdd = seq.numeric().known(BigInteger.ONE);
+      return new IterationState(exponent, accEven, accOdd);
```
pffrandsen: This gives quite unnecessary multiplication of known values.

pffrandsen: I redid your solution to save the multiplications

GuutBoy: Yikes, thanks for catching that!

PR #313, opened 2018-11-07
Matrix subtraction and transposition

jonas-lj:
Add functionality for subtracting and transposing secret shared matrices.


core/src/main/java/dk/alexandra/fresco/lib/real/DefaultLinearAlgebra.java
```diff
@@ -73,6 +73,30 @@ protected DefaultLinearAlgebra(ProtocolBuilderNumeric builder) {
     });
   }
 
+  @Override
+  public DRes<Matrix<DRes<SReal>>> sub(DRes<Matrix<DRes<SReal>>> a, DRes<Matrix<DRes<SReal>>> b) {
+    return builder.par(par -> {
+      return add(par, a.out(), b.out(),
```
GuutBoy: Please find a better name for the `add` method used here. The fact, that it is used to subtract here, suggests that the name is inaccurate :).

core/src/main/java/dk/alexandra/fresco/lib/real/RealLinearAlgebra.java
```diff
@@ -26,6 +26,33 @@
    */
   DRes<Matrix<DRes<SReal>>> add(Matrix<BigDecimal> a, DRes<Matrix<DRes<SReal>>> b);
 
+  /**
+   * Subtracts two secret values and returns the result.
+   *
+   * @param a Secret value 1
```
GuutBoy: This is a bit confusing, the value does not have to do with the number 1. How about `first secret value`/`second secret value`

PR #322, opened 2018-12-03
Adding a Docker file

GuutBoy:
By popular request, this adds a docker file and some basic documentation on how to build and run a FRESCO docker container using it.


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:18.04
+RUN apt-get update -y
+RUN apt-get upgrade -y
+RUN apt-get install -y openjdk-8-jdk maven make
```
pffrandsen: Since the java 11 is the next LTS, maybe let the docker work on that - I am not suggestion using the language features. But rather build and run it on java 11

GuutBoy: Good idea, but I have actually experienced problems building the project with Java 11 (or maybe it was 10?). That’s why I went for 8 here. I will test if 11 works before I do the merge.

PR #323, opened 2018-12-04
Round reduced MiMC

GuutBoy:
A version of MiMC with reduced encryption rounds. Suitable for use as a PRF as described in https://eprint.iacr.org/2016/542.pdf

Also adds functionality to cache a number of constants used in the encryption process.

### DIFF
--- a/core/src/main/java/dk/alexandra/fresco/lib/crypto/mimc/MiMCDecryption.java
+++ b/core/src/main/java/dk/alexandra/fresco/lib/crypto/mimc/MiMCDecryption.java
@@ -54,7 +54,7 @@ public class MiMCDecryption implements Computation<SInt, ProtocolBuilderNumeric>
    * @param requiredRounds The number of rounds to use.
    */
   public MiMCDecryption(DRes<SInt> cipherText, DRes<SInt> encryptionKey, int requiredRounds) {
-    this(cipherText, encryptionKey, USE_DEFAULT_ROUNDS, new MimcConstants());
+    this(cipherText, encryptionKey, requiredRounds, new MimcConstants());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/java/dk/alexandra/fresco/lib/crypto/mimc/MimcConstants.java
```diff
@@ -0,0 +1,50 @@
...
+  private static Map<Pair<Integer, BigInteger>, BigInteger> roundConstants = new HashMap<>();
+
+  private MimcConstants() {
+    // Should not be instantiated
+  }
+
+  /**
+   * Returns a random but deterministic constant depending on the round and the field size. It might
+   * be a better idea to pre-generate the constants using SecureRandom, and distribute them among
+   * the participants if run in production.
```
pffrandsen: This javadoc leads me to believe these constants should not be generated on the fly but rather be supplied to the mimc de-/encryption?

GuutBoy: That is a really good point. I wonder if it was done like this originally because it is dependent on the modulus, which is generally not known at construction.

GuutBoy: Actually, I am not really sure what would be a good way to generate and represent these constants.

GuutBoy:
@pffrandsen how do you like this adjustment?

core/src/main/java/dk/alexandra/fresco/lib/crypto/mimc/MimcConstants.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>
+ * Constants are sampled from a DRNG with seed set to 0 (this should be sufficiently "random" for
+ * MiMC). Each round constant is sampled in order of the round numbers. This should ensure that each
+ * party samples the same round constants in the same order.
+ * </p>
+ */
+public final class MimcConstants implements MimcRoundConstantFactory {
+
+  private List<BigInteger> roundConstants;
+  private Drng drng;
```
pffrandsen: final

pffrandsen: Otherwise fine!

GuutBoy: Cool thanks.

PR #332, opened 2019-01-23
Make last demo processes run in foreground

abspoel:
This makes running the demos a little more visual, imo.


GuutBoy:
I agree, good idea!

PR #333, opened 2019-02-12
Fixing infinite recursion issue with OddEvenMerge when input size is …

basitkhurram:
…greater than four.

Modifying TestOddEvenMerge to test sorting for input of size eight.

Adding a constructor for OddEvenMerge with the `mergePresortedHalves` flag to allow merging
a list with individually sorted halves into a fully sorted list.

Adding a test in TestOddEvenMerge for the `mergePresortedHalves` flag.


basitkhurram:
Looks like there was something wrong with the indexing. I didn't check where exactly the issue was, but matching the indexing from [Flensburg](http://www.inf.fh-flensburg.de/lang/algorithmen/sortieren/networks/oemen.htm) appears to have fixed the issue.

PR #340, opened 2019-03-29
Faster log

jonas-lj:
Faster logarithms for fixed point numbers


n1v0lg:
Currently the result precision is hard-coded to 8 in the tests.

Is there a manageable way to guarantee result precision that matches our fixed point precision value instead?

jonas-lj:
You're right. However, the precision depends strongly on the input. I could write some documentation about the expected precision depending on the input size?

n1v0lg:
> You're right. However, the precision depends strongly on the input. I could write some documentation about the expected precision depending on the input size?

Yep, that would be great.

jonas-lj:
I added a lengthy and rather mathy javadoc explaining the error and how it can be compensated for for large inputs.

Issue #324, opened 2018-12-10
Bug in MAC check for SPDZ-based protocols


n1v0lg:
SPDZ, Mascot, and SPDZ2k require coin tossing to run every time a MAC check is executed.

Currently, we only run coin tossing once and re-use the resulting DRBG for all MAC checks. This is not secure and should be fixed for:

- [ ] SPDZ
- [ ] Mascot
- [ ] SPDZ2k

jot2re:
Should be fixable by running coin-tossing each time there is a MAC check.

PR #343, opened 2019-06-07
324 mac check spdz

n1v0lg:
This PR addresses #324 which is a bug in our SPDZ mac-checking protocol.

In particular, we need to run a fresh coin tossing each time we run a MAC check. To implement this, the PR moves several generic computations (for instance `BroadcastComputation`) out of SPDZ2k and into core, so these can be re-used across protocol suites.

Re-using these generic computations also allows us to remove some classes that were previously SPDZ-specific (e.g., `SpdzCommitProtocol`) and use the same commitment functionality as is currently used in MASCOT.


core/src/main/java/dk/alexandra/fresco/lib/generic/BroadcastComputation.java
```diff
@@ -17,13 +15,19 @@
...
-  BroadcastComputation(List<byte[]> input) {
+  public BroadcastComputation(List<byte[]> input, boolean runValidation) {
     this.input = input;
+    this.runValidation = runValidation;
   }
 
-  BroadcastComputation(byte[] input) {
-    this(java.util.Collections.singletonList(input));
+  public BroadcastComputation(byte[] input, boolean runValidation) {
+    this(java.util.Collections.singletonList(input), runValidation);
```
rubensayshi: while you're touching this line might as well get rid of the long class name `java.util.Collections`

core/src/main/java/dk/alexandra/fresco/lib/generic/CommitmentComputation.java
```diff
@@ -33,12 +33,28 @@ public Spdz2kCommitmentComputation(ByteSerializer<HashBasedCommitment> commitmen
   public DRes<List<byte[]>> buildComputation(ProtocolBuilderNumeric builder) {
     HashBasedCommitment ownCommitment = new HashBasedCommitment();
     byte[] ownOpening = ownCommitment.commit(localDrbg, value);
-    return builder.seq(new BroadcastComputation<>(commitmentSerializer.serialize(ownCommitment)))
+    return builder.seq(
+        seq -> {
+          if (noOfParties > 2) {
```
rubensayshi: would be nice to comment here that for <= 2 `InsecureBroadcastProtocol` can be used because there's no other parties to verify against anyway. (I know it "makes sense", but it doesn't hurt to explain)

core/src/test/java/dk/alexandra/fresco/lib/generic/MaliciousCommitmentComputation.java
```diff
@@ -0,0 +1,53 @@
...
+    this.localDrbg = localDrbg;
+  }
+
+  @Override
+  public DRes<List<byte[]>> buildComputation(ProtocolBuilderNumeric builder) {
+    HashBasedCommitment ownCommitment = new HashBasedCommitment();
+    byte[] ownOpening = ownCommitment.commit(localDrbg, value);
+    return builder.seq(
+        seq -> {
+          if (noOfParties > 2) {
```
rubensayshi: same as CommitmentComputation.java, would be nice to comment why it's different for <= 2

core/src/main/java/dk/alexandra/fresco/lib/generic/CommitmentComputation.java
```diff
@@ -1,27 +1,27 @@
...
- * Protocol for all parties to commit to a value and open it to the other parties.
+ * Protocol for all parties to commit to a value each and open it to the other parties.
  */
-public class Spdz2kCommitmentComputation implements
+public class CommitmentComputation implements
     Computation<List<byte[]>, ProtocolBuilderNumeric> {
 
   private final ByteSerializer<HashBasedCommitment> commitmentSerializer;
   private final byte[] value;
   private final int noOfParties;
```
rubensayshi: you can retrieve the `noOfParties` from the `builder` in `buildComputation`  so you don't need to pass it into the constructor, same for `CoinTossingComputation` (which only needs it to pass it into the constructor of `CommitmentComputation`)

```java
final int noOfParties = builder.getBasicNumericContext().getNoOfParties();
```

n1v0lg:
@rubensayshi awesome, thanks for the review! All good points. Addressed in 5f9eb04.

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -1,64 +1,73 @@
...
+    this.jointDrbgSupplier = jointDrbgSupplier;
     this.alpha = alpha;
+    this.drbgSeedBitLength = drbgSeedBitLength;
   }
 
   @Override
   public DRes<Void> buildComputation(ProtocolBuilderNumeric builder) {
-    return builder
-        .seq(seq -> {
+    final int noOfParties = builder.getBasicNumericContext().getNoOfParties();
```
jot2re: noOfParties is never used

suite/spdz2k/src/main/java/dk/alexandra/fresco/suite/spdz2k/protocols/computations/Spdz2kMacCheckComputation.java
```diff
@@ -74,7 +76,7 @@ public Spdz2kMacCheckComputation(Pair<List<Spdz2kSInt<PlainT>>, List<PlainT>> to
             List<byte[]> sharesLowBits = authenticatedElements.stream()
                 .map(element -> element.getShare().getLeastSignificant().toByteArray())
                 .collect(Collectors.toList());
-            return new BroadcastComputation<ProtocolBuilderNumeric>(sharesLowBits)
+            return new BroadcastComputation<ProtocolBuilderNumeric>(sharesLowBits, true)
```
jot2re: Maybe I am missing something but won't we also need to sample the coefficients for the MAC check in SPDZ2k as for SPDZ during the computation of the MAC check? In that case it seems that it is not enough to simply call sampleCoefficients in the constructor but it should also be part of the buildComputation.

suite/spdz2k/src/main/java/dk/alexandra/fresco/suite/spdz2k/resource/Spdz2kResourcePoolImpl.java
```diff
@@ -92,7 +92,7 @@ public void initializeJointRandomness(Supplier<Network> networkSupplier,
     Network network = networkSupplier.get();
     Computation<byte[], ProtocolBuilderNumeric> coinTossing =
         new CoinTossingComputation(seedLength, new HashBasedCommitmentSerializer(),
-            getNoOfParties(), getLocalRandomGenerator());
+            getLocalRandomGenerator());
```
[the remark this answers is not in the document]
n1v0lg: Ah yes, nice catch! Sorry, should have been more specific in my PR description. This PR **only** fixes SPDZ, it does not yet fix SPDZ2k. The reason code in the `spdz2k` module changed at all in this PR is that there was some generic functionality (e.g., Broadcast) that I moved over to `core`. I'm planning to fix SPDZ2k and Mascot in separate PRs. Sound good?

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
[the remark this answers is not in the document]
n1v0lg: Fixed in 626fbfc.

core/src/main/java/dk/alexandra/fresco/lib/generic/BroadcastComputation.java
```diff
@@ -17,13 +16,19 @@
     Computation<List<byte[]>, BuilderT> {
 
   private final List<byte[]> input;
+  private final boolean runValidation;
 
-  BroadcastComputation(List<byte[]> input) {
+  public BroadcastComputation(List<byte[]> input, boolean runValidation) {
```
pffrandsen: I think it is reasonable to document this class and the constructor - especially when it it resonable to use one or the other

suite/spdz/src/main/java/dk/alexandra/fresco/suite/spdz/gates/SpdzMacCheckProtocol.java
```diff
@@ -1,64 +1,72 @@
...
   }
 
   @Override
   public DRes<Void> buildComputation(ProtocolBuilderNumeric builder) {
-    return builder
-        .seq(seq -> {
+    final AesCtrDrbg localDrbg = new AesCtrDrbg();
+    final HashBasedCommitmentSerializer commitmentSerializer = new HashBasedCommitmentSerializer();
+
+    return builder.seq(new CoinTossingComputation(drbgSeedBitLength / 8,
```
pffrandsen: Please reconsider the alignment here

pffrandsen: Actually check the entire file :)

suite/spdz2k/src/main/java/dk/alexandra/fresco/suite/spdz2k/resource/Spdz2kResourcePoolImpl.java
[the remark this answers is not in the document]
jot2re: Yes, sounds good. Thanks for clarifying.

PR #345, opened 2019-07-09
Simplify fixed functions

jonas-lj:
More predictable and precise transcendental function, namely exp, sqrt and log. Also introduces some new functionality.

This pull request contains pull request 344, so if this is accepted, the other can be ignored.


core/src/test/java/dk/alexandra/fresco/lib/real/MathTests.java
```diff
@@ -342,4 +339,166 @@ public void test() throws Exception {
       };
     }
   }
+
+  public static class TestReciprocal<ResourcePoolT extends ResourcePool>
+      extends TestThreadFactory<ResourcePoolT, ProtocolBuilderNumeric> {
+
+    @Override
+    public TestThread<ResourcePoolT, ProtocolBuilderNumeric> next() {
+      List<BigDecimal> openInputs = Stream.of(.01, 0.1, 1.1, 2.1, 3.1, 4.1, 5.1, 10.1)
```
pffrandsen: Maybe create a shared vault of test data for these (and other) real tests - they seem to be chosen somewhat biased?

core/src/test/java/dk/alexandra/fresco/lib/real/MathTests.java
```diff
@@ -342,4 +339,166 @@ public void test() throws Exception {
...
+  public static class TestConstantPolynomial<ResourcePoolT extends ResourcePool>
+      extends TestThreadFactory<ResourcePoolT, ProtocolBuilderNumeric> {
+
+    @Override
+    public TestThread<ResourcePoolT, ProtocolBuilderNumeric> next() {
+      List<BigDecimal> openInputs =
+          Stream.of(1.0, 2.0, 3.0, 10.0)
+              .map(BigDecimal::valueOf).collect(Collectors.toList());
+
+              double[] p = new double[] {7.0};
```
pffrandsen: indent?

core/src/test/java/dk/alexandra/fresco/lib/real/MathTests.java
jonas-lj: Maybe. But the functions has different restrictions on input (size, sign, etc.), so I'm not sure how to do this in a nice way. For now, I have changed the test data so it's obvious that it was not generated randomly, but rather chosen to represent input of different order of magnitudes.

PR #349, opened 2019-09-25
Fast reduction

jonas-lj:
Fast modular reduction using Barrett Reduction. Is about 28% than BigInteger::mod.


core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/field/BigIntegerModulus.java
```diff
@@ -29,7 +32,19 @@
...
+  
+  /**
+   * Compute <i>x</i> modulus this modulus <i>M</i> assuming that <i>0 &le; x < M<sup>2</sup></i>
+   * 
+   * @param x A non-negative integer smaller than the modulus squared.
+   * @return A BigInteger equal to x mod M.
+   */
+  BigInteger reduceModThis(BigInteger x) {
+    // We don't serialize the reducer and create it lazily if needed
+    reducer = Objects.requireNonNullElse(reducer, new ModularReducer(value));
```
n1v0lg: This is not Java 8 compatible.

core/src/test/java/dk/alexandra/fresco/framework/util/TestModularReduction.java
```diff
@@ -0,0 +1,34 @@
+package dk.alexandra.fresco.framework.util;
+
+import static org.junit.Assert.assertEquals;
+
+import java.math.BigInteger;
+
+import org.junit.Test;
+
+public class TestModularReduction {
```
n1v0lg: We should at a minimum test for:

* Each modulus we plan on using in SPDZ (i.e., those in ModulusFinder)
* Non-negative inputs smaller than the modulus
* Edge-cases: zero input, input equal to modulus

There are probably more edge cases, but I haven't looked through the implementation yet.

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/field/BigIntegerModulus.java
jonas-lj: Fixed

core/src/test/java/dk/alexandra/fresco/framework/util/TestModularReduction.java
jonas-lj: I extended the tests (which caught an error :))

core/src/main/java/dk/alexandra/fresco/framework/builder/numeric/field/BigIntegerModulus.java
```diff
@@ -29,7 +32,21 @@
...
+  
+  /**
+   * Compute <i>x</i> modulus this modulus <i>M</i> assuming that <i>0 &le; x < M<sup>2</sup></i>
+   * 
+   * @param x A non-negative integer smaller than the modulus squared.
+   * @return A BigInteger equal to x mod M.
+   */
+  BigInteger reduceModThis(BigInteger x) {
+    // We don't serialize the reducer and create it lazily if needed
+    if (Objects.isNull(reducer)) {
```
pffrandsen: why not just == null?

jonas-lj: No reason. == null might be slightly more readable though, so I'll change it

PR #351, opened 2019-10-24
Mascot demo

jonas-lj:
Add option for using preprocessing using MASCOT in demos


demos/common/src/main/java/dk/alexandra/fresco/demo/cli/CmdLineProtocolSuite.java
```diff
@@ -119,27 +141,121 @@ private Properties getProperties(Properties properties) {
...
-    }
-    if (strategy == PreprocessingStrategy.STATIC) {
-      int noOfThreadsUsed = 1;
-      String storageName =
-          SpdzStorageDataSupplier.STORAGE_NAME_PREFIX + noOfThreadsUsed + "_" + myId + "_" + 0
-              + "_";
-      supplier = new SpdzStorageDataSupplier(
-          new FilebasedStreamedStorageImpl(new InMemoryStorage()), storageName, noOfPlayers);
+
+    switch (strategy) {
```
pffrandsen: switch is very hard to test - that could explain some of the loss of coverage.

pffrandsen:
Note that the README for equals demo needs updating. It looks like a copy

demos/common/src/main/java/dk/alexandra/fresco/demo/cli/CmdLineProtocolSuite.java
jonas-lj: That, and the exp-pipe generation is not tested. Hence the "equals demo", which is now moved to the demos-common in order to get 100% codecov in that project.

PR #356, opened 2020-08-26
Move truncate and normalizesint

jonas-lj:
Truncate and NormalizeSInt are used in the fixed point lib, but may have usages elsewhere. This PR allows access to them through the advanced numeric builder.


jonas-lj:
This PR includes PR #355, so that should be merged first or deleted after this has been merged.

pffrandsen (approved):

PR #364, opened 2020-09-16
Kd move libraries

kdath:


core/src/main/java/dk/alexandra/fresco/logging/NumericSuiteLogging.java
```diff
@@ -30,8 +26,14 @@
    * @param protocolSuite the original protocol suite to log for.
    */
   public NumericSuiteLogging(ProtocolSuiteNumeric<ResourcePoolT> protocolSuite) {
+    this(protocolSuite, new PerformanceLoggerCountingAggregate());
+  }
+
+  protected NumericSuiteLogging(ProtocolSuiteNumeric<ResourcePoolT> protocolSuite,
```
pffrandsen, line 32: why not package private?

lib/common/src/main/java/dk/alexandra/fresco/lib/common/collections/Collections.java
```diff
@@ -1,17 +1,28 @@
...
-import dk.alexandra.fresco.framework.util.RowPairD;
 import dk.alexandra.fresco.framework.value.SInt;
-import dk.alexandra.fresco.lib.collections.Matrix;
+import dk.alexandra.fresco.lib.common.math.DefaultAdvancedNumeric;
+import dk.alexandra.fresco.lib.common.util.RowPairD;
 import java.math.BigInteger;
 import java.util.List;
 
 /** Interface for operations on collections. This also includes opening and closing values. */
 public interface Collections extends ComputationDirectory {
```
pffrandsen, line 14: I realize this is not a part of the merge, but the name is unfortunate.

Maybe call it MatrixCol or similar, Collections reside already in java....

core/src/main/java/dk/alexandra/fresco/lib/field/integer/BasicNumericContext.java
```diff
@@ -21,13 +22,15 @@
...
+   * @param precision the fixed point precision when using the fixed point library
    */
   public BasicNumericContext(int maxBitLength, int myId, int noOfParties,
-      FieldDefinition fieldDefinition) {
+      FieldDefinition fieldDefinition, int precision) {
     this.maxBitLength = maxBitLength;
     this.myId = myId;
     this.noOfParties = noOfParties;
     this.fieldDefinition = fieldDefinition;
+    this.precision = precision;
```
pffrandsen, line 33: how can precision be defined here - without the F.P. library residing together? Most likely similar to maxBitLength?

pffrandsen (approved):
This is huge, quite hard to check everything. I have added a few comments from reading the most central classes

core/src/main/java/dk/alexandra/fresco/lib/field/integer/BasicNumericContext.java
jonas-lj: That's true. It's to avoid having to specify the precision every time a FixedNumeric is created. The only solution I can come up with is to put a default precision in DefaultFixedNumeric.

pffrandsen: Maybe call it that? defaultPrecision?

jonas-lj: Do you mean we should keep it in BasicNumericContext and just rename it, or put a constant somewhere in the fixed-lib?

pffrandsen: Well, I would prefer to keep the magic constants close to their use, however if it too complicated, then rename precision to defaultFixedPointPrecision

jonas-lj: I think that's a descent solution for now. It breaks the separation of modules but makes it easier to understand what the number actually means

PR #370, opened 2021-01-19
Optimised modular inverse for pseudo Mersenne primes

jonas-lj:


pffrandsen (changes requested):
Nice

Consider adding all the comments for trackaibility from the article?

pffrandsen (approved):

PR #373, opened 2021-03-11
Bit triples

kdath:
Adding protocol for producing malicious secure bit triples, described in https://eprint.iacr.org/2015/901.pdf.


tools/bitTriples/src/test/java/dk/alexandra/fresco/tools/bitTriples/BitTriplesTestContext.java
```diff
@@ -0,0 +1,70 @@
...
+  private final Network network;
+
+  /**
+   * Creates new test context.
+   */
+  public BitTriplesTestContext(int myId, int noOfParties, int instanceId,
+      BitTripleSecurityParameters securityParameters) {
+    this.network = new SocketNetwork(defaultNetworkConfiguration(myId, noOfParties));
+    byte[] drbgSeed = new byte[securityParameters.getPrgSeedBitLength() / 8];
+    new Random(myId).nextBytes(drbgSeed); // Parties need to have same seed.
```
ellieMayVelasquez, line 32: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once).
This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.
If it is important that the generated Random numbers not be guessable, you *must* not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).

GuutBoy: @ellieMayVelasquez This is a test context though. I don't think it matters too much.

ellieMayVelasquez: thanks for your reply. noticed :)

PR #379, opened 2021-09-01
Access to values in SpdzSInt and FieldElement. Closable on SecureComputationEngine.

eoasecata:


GuutBoy:
Come on! Could you please comment your PR's a bit more? It's really hard to follow what is going on this way 😛

jonas-lj:
Sure! We did all communication about this PR through other channels, but I see why that doesn't really help you :). Partisia requested some convenience methods to make it easier to extract shares from secret shared values, eg. extracting BigIntegers. They also needed SecureComputationEngine and Mascot to be Closable -- my guess is because they run multiple instances of these and not just a single one which starts and stops with the application as in the existing demos.

PR #384, opened 2021-11-01
Use norm to compute bit length

jonas-lj:
In order to compute the bit length, fresco computes the entire bit string from right shifts. However, a more efficient method to do this is implemented to support fixed point arithmetic, so this pull request suggests that this more efficient method is also used to compute the bit length.


pffrandsen (approved):
nice!

PR #389, opened 2022-03-05
Ot base implementation from Chou Orlandi with fixes

### LINKED ISSUE #388: HL17 baseOt implementation
Naor Pinkas is not the fastest base OT. 
By changing the protocol to [HL17 ](https://eprint.iacr.org/2017/1011) we can cut speedup the offline phase. (From 31 to 17 seconds in my tests, when using BigInteger). (In the case of Spdz with Mascot)
The HL17 is based on ChouOrland OT

Gugi264:
Closes #388


jot2re:
As discussed in the comments on issue #388 I unfortunately have to reject the PR.

jot2re:
After more discussion with Claudio Orlandi, I realised that the amount of changes to make the protocol secure are really minimal. Thus I think it still makes sense to merge it.
I am really sorry for being too quick to reject the PR previously. 
I understand that you probably don't have time to implement the suggestions at this point. If not, I will probably just do it myself.

Gugi264:
Hi, what are the necessary changes? If they are really only small I can implement them.

Gugi264:
Ok :)

jot2re:
@quacktiamauct can you implement the changes I have requested?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -94,27 +95,24 @@ public StrictBitVector receive(boolean choiceBit) {
...
+        T U;
         if (choiceBit == false) {
-            R = T.exponentiation(BigInteger.ZERO).groupOp(getGenerator().exponentiation(x));
+            U = A.exponentiation(BigInteger.ZERO).groupOp(getGenerator().exponentiation(x));
         } else {
-            R = T.exponentiation(BigInteger.ONE).groupOp(getGenerator().exponentiation(x));
+            U = A.exponentiation(BigInteger.ONE).groupOp(getGenerator().exponentiation(x));
         }
-        network.send(otherId, R.toByteArray());
+        network.send(otherId, U.toByteArray());
```
jot2re, line 110: We still need to ensure that `toByteArray` always returns an array of the same amount of bytes regardless of the size of the underlying element. If not and `toByteArray` returns things as big endian we end up in situation where two different values will have the same byte representation.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/ECNaorPinkasOt.java
```diff
@@ -0,0 +1,126 @@
...
+    int bytesNeeded = (2 * curve.getFieldSize()+7) / 8;
+    byte[] nextInt = new byte[bytesNeeded];
+    rand.nextBytes(nextInt);
+    BigInteger rawInt = new BigInteger(1, nextInt);
+    return curve.fromBigInteger(rawInt.mod(curve.getField().getCharacteristic()));
+  }
+
+  @Override
+  BouncyCastleECCElement decodeElement(byte[] bytes) {
+    return new BouncyCastleECCElement(this.curve.decodePoint(bytes));
```
jot2re: Ensure that the element is of the correct order (when the curve does not have cofactor 1). I.e. that the point is actually valid and that the point multiplied by the cofactor is different from the point at infinity.
It might also be relevant to check that it is not on the point of infinity.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -128,28 +126,22 @@ private byte[] receiveRandomOt(boolean choiceBit) {
...
         k0Hash = hashDigest.digest();
 
-        hashDigest.update(S.toByteArray());
-        hashDigest.update(R.toByteArray());
-        // R^y*T^(-jy) with j = 1:
-        // R^y * T^(-1y) == (R + (-T)) * y
-        hashDigest.update((R.groupOp(T.inverse())).exponentiation(y).toByteArray());
+        hashDigest.update(A.toByteArray());
+        // (U * B^(-1))^y
+        hashDigest.update((U.groupOp(A.inverse())).exponentiation(y).toByteArray());
```
jot2re, line 144: In the following SHAKE is used.
Why use both SHA256 and SHAKE? Would it not make sense to use the same hash function?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/ECNaorPinkasOt.java
jot2re: This is also an issue in `ECChouOrlandi` when decoding

jot2re (changes requested):
There are still a lot of security issues from the previous review that has not been addressed. 
I have fixed some of them and explicitly written down the remaining issues and how to fix them

Quacktiamauct:
Fixed in https://github.com/aicis/fresco/pull/389/commits/2ce267e45e817b797bdeff2599c5c20f5de294c6

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
Quacktiamauct: The underlying BigInteger implementation always returns a big endian representation. The EC implementation an encoded non-compressed point. The first one has no guarantees in the term of byte length, but the second will depend directly on the field size. However I am not quite sure how a variable byte length would be a problem? In any case should we then pad the sent bytes?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/ECNaorPinkasOt.java
Quacktiamauct: BouncyCastle already validates points when decoding, both that they are on the curve and that the order time.
It runs the following:
```java
    protected boolean satisfiesOrder()
    {
        if (ECConstants.ONE.equals(curve.getCofactor()))
        {
            return true;
        }

        BigInteger n = curve.getOrder();

        // TODO Require order to be available for all curves

        return n == null || ECAlgorithms.referenceMultiply(this, n).isInfinity();
    }
```
from `org.bouncycastle.math.ec.ECPoint`

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
Quacktiamauct: I'm not sure I am following, the later `hashToFieldElement` which uses SHAKE is never called, and the highlighted code only uses SHA-256 for hashing?

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -0,0 +1,187 @@
...
+import java.security.MessageDigest;
+
+import static org.bouncycastle.pqc.math.linearalgebra.BigEndianConversions.I2OSP;
+
+/**
+ * Uses Chou-Orlandi with fixes as seen in https://eprint.iacr.org/2017/1011
+ */
+public abstract class AbstractChouOrlandiOT<T extends InterfaceOtElement<T>> implements Ot {
+
+    private static final String HASH_ALGORITHM = "SHA-256";
```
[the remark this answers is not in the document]
Quacktiamauct: Solved in 0c254516f

demos/distance/src/test/java/dk/alexandra/fresco/demo/TestDistanceDemo.java
```diff
@@ -128,7 +128,7 @@ public void testDistanceFromCmdLine3Party() throws Exception {
       try {
         DistanceDemo.main(
             new String[]{"-i", "1", "-p", "1:localhost:8081", "-p", "2:localhost:8082",
-                "-p", "3:localhost:8083", "-s", "dummyArithmetic", "-x", "10", "-y", "10"});
+                "-p", "3:localhost:8083", "-s", "spdz", "-x", "10", "-y", "10"});
```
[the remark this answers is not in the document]
Quacktiamauct: Solved in 4fb38fa142f7e4ed5679bd3fba59a720b0956e28

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -0,0 +1,187 @@
...
+     * @return The two random messages sent by the sender.
+     */
+    private Pair<byte[], byte[]> sendRandomOt() {
+        // y
+        BigInteger y = randNum.nextBigInteger(getDhModulus());
+        // S
+        T S = this.getGenerator().exponentiation(y);
+        network.send(otherId, S.toByteArray());
+        // T = G(S)
+        T T = hashToFieldElement(S, "FRESCO|ChouOrlandi|OT");
```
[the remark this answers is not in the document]
Quacktiamauct: Fixed in 4733c6681ff9b17a9080b8d278c2e914c08e553d

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/BouncyCastleChouOrlandi.java
```diff
@@ -0,0 +1,55 @@
...
+        Security.addProvider(new BouncyCastleProvider());
+        X9ECParameters ecP = CustomNamedCurves.getByName("curve25519");
+        this.curve = ecP.getCurve();
+        this.dhModulus = curve.getOrder();
+        this.dhGenerator = ecP.getG();
+    }
+
+    @Override
+    BouncyCastleECCElement decodeElement(byte[] bytes) {
+        return new BouncyCastleECCElement(this.curve.decodePoint(bytes));
```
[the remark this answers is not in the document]
Quacktiamauct: BC does both of these, except the infinity check.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -0,0 +1,187 @@
...
+        if (choiceBit == false) {
+            R = T.exponentiation(BigInteger.ZERO).groupOp(getGenerator().exponentiation(x));
+        } else {
+            R = T.exponentiation(BigInteger.ONE).groupOp(getGenerator().exponentiation(x));
+        }
+        network.send(otherId, R.toByteArray());
+
+        // k = H(S, R, S^x)
+        byte[] key;
+        hashDigest.update(S.toByteArray());
```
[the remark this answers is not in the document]
Quacktiamauct: I'm not sure I understand by hashing locally?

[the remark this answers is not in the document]
jot2re: Sorry, badly written from my side. The we need to ensure that what gets returned by `toByteArray` is always the same amount of bytes, regardless of the magnitude of the number it represents. Since it is most likely big endian this is a problem, and thus the returned value need to be prepended by empty bytes if S is not a number that is large enough to use the maximum amount of bytes

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
```diff
@@ -0,0 +1,187 @@
...
+        T S = this.decodeElement(network.receive(otherId));
+        // T = G(S)
+        T T = hashToFieldElement(S, "FRESCO|ChouOrlandi|OT");
+        // R = T^c*g^x
+        // if c = 0 -> R = g^x
+        // if c = 1 -> R = T*g^x
+        BigInteger x = randNum.nextBigInteger(getDhModulus());
+        T R;
+        if (choiceBit == false) {
+            R = T.exponentiation(BigInteger.ZERO).groupOp(getGenerator().exponentiation(x));
```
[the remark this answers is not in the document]
jot2re: Yes, we currently don't care about timing attacks. In particular since the code is Java which is naturally vulnerable to such things

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/BouncyCastleChouOrlandi.java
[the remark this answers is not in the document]
jot2re: Either also validate manually or write a comment that it is expected from the library. It is not given that such things are validated when decoding a point and we cannot be sure bouncy castle does not change this functionality in the future. A redundant check is always better than a missing check

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
jot2re: Yes, padding should be added to ensuring equal byte length for all numbers

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/ECNaorPinkasOt.java
jot2re: As mentioned above ideally manually check this as well or write a comment that it is expected from the library. But the functionality in the library might change, hence it is better to also check it manually since this can _completely_ unmine the security if not done

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
jot2re: If SHAKE is never used then the code should be removed. Currently there is code using both kinds, but if the SHAKE code is now dead it should just be removed

Quacktiamauct: Resolved by 600750daa

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
Quacktiamauct: Actually, I'm not sure this is a problem since the byte-representation is prepended with the sign in the first byte.

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
Quacktiamauct: Solved by 20615343b

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
jot2re: You are right that for network communication it does not matter how many bytes is used.
However when we need to use the result we can get into issues if we  have a number that is represented using a 31 byte big endian integer, and we need to hash this. Thus a bunch of 0-bytes gets appended. Now the result will represent a much larger number than it actually is. In that case the number should be prepended with 0-bytes to ensure consistent encoding once it gets mapped to a 32 byte array (or something larger)

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/ECNaorPinkasOt.java
Quacktiamauct: Added a comment in 0375fd3

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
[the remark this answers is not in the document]
Quacktiamauct: Solved by 0375fd3

tools/ot/src/main/java/dk/alexandra/fresco/tools/ot/base/AbstractChouOrlandiOT.java
Quacktiamauct: Solved by 0375fd3

jot2re (changes requested):
Look good. I think the only issue is that there is not full coverage with the current tests. If you could add a couple of tests to ensure this (so the CI/CD won't complain once it gets merged) that would be great

jot2re (approved):

PR #393, opened 2022-06-09
Enable custom SocketFactory and ServerSocketFactory in SocketNetwork

jonas-lj:


GuutBoy:
This seems silly, you could already do this by supplying a socketmap from the appropriate connector.

jonas-lj:
> This seems silly, you could already do this by supplying a socketmap from the appropriate connector.

That's true. It was a specific request for a convenience constructor from a user, but it's true that calling the constructor with a given connector will give the same result.

PR #401, opened 2022-09-14
Fix CI breaking in Forks

Quacktiamauct:
Fixes CI breaking for forks due to insufficient permissions.

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  checks: write
+  contents: write
+  pull-requests: write
 
 jobs:
   build:


jot2re (changes requested):
Are you sure such extensive permissions are needed? 
I don't really know about GitHub actions, but I know that on another project where we are doing the same (code coverage report), it is not needed to set write permissions.
Maybe those yml files can help? https://github.com/TokenScript/attestation/tree/main/.github/workflows

Quacktiamauct:
> Are you sure such extensive permissions are needed?
> I don't really know about GitHub actions, but I know that on another project where we are doing the same (code coverage report), it is not needed to set write permissions.
> Maybe those yml files can help? https://github.com/TokenScript/attestation/tree/main/.github/workflows

I'm not sure, however it seems to be on by default  (see [here](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token)) except in the cases of forks , which is why it fails in the other PR.
The solution I applied here is suggested from the issue [#test-reporter/issues/168](https://github.com/dorny/test-reporter/issues/168), however I am not sure if the suggested fix works in our case.

The issue might be solved upstream instead, so we could also disable the test-report either partially for forks or all together.

jot2re (approved):

PR #411, opened 2023-09-28
Socketnetwork timeout and polling

JeppeOvervad:


core/src/main/java/dk/alexandra/fresco/framework/network/socket/Connector.java
```diff
@@ -137,7 +137,7 @@ private Map<Integer, Socket> connectClient(final NetworkConfiguration conf)
           } catch (ConnectException e) {
             // A ConnectionException is expected if the opposing side is not listening for our
             // connection attempt yet. We ignore this and try again.
-            Thread.sleep(1L << ++attempts);
+            Thread.sleep(Math.min(1 << attempts++, 5_000));
```
JeppeOvervad, line 140: The choice of an upper bound is based on the assumption that we always want to attempt to connect frequently - and 5 seconds was picked based on our own observations. 

Alternatively this upper bound could be made configurable, or the entire calculation could be configurable - e.g.  some `Function<Integer, Long> pollInterval`  with the default being `attempts -> 1 << attemps`

Quacktiamauct: It should probably just keep waiting the same amount, scale linearly, or max out after a period. We probably never want to wait 24 days (2^31 ms) between retries. But the existing implementation is a bit weird. 5 seconds is probably a good amount.

JeppeOvervad:
I assume the build failure is due to missing permissions on the job

https://github.com/aicis/fresco/blob/master/.github/workflows/test-report.yml ? 

(Similar to https://github.com/aicis/fresco/commit/80c3c9ba24ab6d5f44fdc7c73460f931aa70322b ? )

Quacktiamauct:
Yea, GitHub actions have a tendency to break with forks. Anyway the tests seem to run fine locally and the new constructor seems like a good candidate.

Quacktiamauct (approved):

PR #433, opened 2025-02-26
Add party id to hash based commitment to avoid any adversary using the commitment

Thomar10:


Thomar10:
@quackzar Anything needed from me to push through?

Thomar10:
@quackzar Seems like the pipeline failed. Let me know if I should do anything

quackzar:
It's just the test-reporter action that is broken. The test suite runs fine now.

